PR #410, opened 2015-02-11
Battle Pet Levelup guide

rpotor:
New guide proposal


Emmaleah:
Query... isn't the wow pro frame hidden while in a pet battle?

rpotor:
Normally it is, but if we do

if event == "PET_BATTLE_OPENING_START" and (not WoWPro.Hidden) and
WoWProCharDB.
AutoHideInsideInstances == true then

instead of

if event == "PET_BATTLE_OPENING_START" and (not WoWPro.Hidden) then

in

WoWPro_Events.lua

then it will show if Auto Hide is unchecked in options.

Shall I send in a modification proposal to that effect or is there some
more efficient way to do this?


2015-02-11 20:07 GMT+01:00 Emmaleah <notifications@github.com>:

> Query... isn't the wow pro frame hidden while in a pet battle?

Ludovicus:
I will add an attribute to the guide that will disable pet battle hiding.

Emmaleah:
What does that entail... or more important to me atm, not hiding in
dungeons. The horde lunar festival guide is auto hiding and I had to
uncheck the enable hiding box.
Btw I have updated the horde lunar festival guide and will post it tonight
or tomorrow as soon as I manage to take a 2nd toon thru.

Ludovicus:
Three things:

1)  I will add a command WoWPro:GuidePetBattle(guide) so a guide can declare itself to not hide during pet battles.
2) Any guide with a step with a Z step in an instance is already NOT supposed to hide.  That is a bug that shows up with the lunar festival guide.
3) I’ll be updating the addon on Tuesday for 6.1 .

Ludovicus:
I profoundly apologize for the delay.

PR #437, opened 2016-09-28
Update 110_110_Blanc_Suramar.lua

itsjamie:
Update coordinates for Leyline Feed objective


Ludovicus:
The primary source is on the website.  I'll transfer your changes.

Ludovicus:
Hmm, I bet there is a hidden floor that resulted in the differences in the coordinates.
Can you pleased go back to that cave and do "/wp where"?

PR #441, opened 2016-10-10
Pet Battle World Quests Initial commit

rpotor:
Guide for pet battle world quests - in progress, not all pet world quests are included yet, but it's useable, about two-thirds of all quests are done.


Ludovicus:
Note:  I will make the changes to add [ability=xxx] tags.

PR #444, opened 2016-12-14
Auto turn-in/accept fixes.

cameljon:
Issues: 
- Quests were not being auto completed or turn-in (consistently). 
- QIDs were not getting updated after first iterations of QUEST_DETAILS, QUEST_PROGRESS, QUEST_GREETING, and GOSSIP_SHOW. 
- Offered QID no longer matched current QID.

Fixes:
- Added CompleteStep to QUEST_DETAILS. Fixed the QID update for that event. (sometimes)
- Wrapped offending events in C_Timer.After for 0.5. The next quest event was getting triggered prior to AutoCompleteQuestUpdate updating QID.

Still broke:
- Very rarely Blizzard calls QUEST_PROGRESS for completing a quest. This isn't handled by WP because it expects there to be a quest reward?
- A delay of 0.5 may be longer than needed. A fast clicker may click quests UI elements before WP handles them. But the events should still fire correctly?


cameljon:
I removed this pull request due to the opposition to the C_Timer function. However, this bug still exists and this is what I've personally been using to make auto-complete/turn-in operational for multiple accept/turn-ins from the same NPC.

It still rarely fails, normally when needing to choose a quest reward and a follow-on quest being offered.

I'm only reopening this to offer a fix for whoever wants it, and to acknowledge the bug exists.

Ludovicus:
Hmm.   The Blizzard UI now uses C_Timer.After all over the place and C_Timer has grown and been used and documented.  So my original objections are no longer relevant.

I'm going to cut another small patch release soon, and will then merge this change and play with it for the following release.

cameljon:
Some guides have a quest much later in the guide that incorrectly gets compared to an offered quest after the QUEST_COMPLETE event is processed. (For all quests in that guide up until the offending quest is completed.) I'll post more debug info when I get home.

`Turn in quest -> NPC offers new quest
WP uses wrong qidx during the new accept quest dialog
Closing dialog and reopening works correctly`

The quest causing this in Highmountain is [The Backdoor] id: 39579.

cameljon:
OK, that fixed it. QUEST_DETAIL just needed a longer delay. The delays may need adjusting depending on ping, or some longer fail safe value?

Ludovicus:
I’ll have to do a debugging run with full event tracing turned on.
There may be another event to trigger off of that is safer.


On Aug 3, 2017, at 12:25 PM, Jon Volden <notifications@github.com<mailto:notifications@github.com>> wrote:


OK, that fixed it. QUEST_DETAIL just needed a longer delay. The delays may need adjusting depending on ping, or some longer fail safe value?

cameljon:
No longer needed.

PR #445, opened 2017-08-12
Quest line removed

eTzmNcbkrng:
This quest line ws removed in 7.1.5. Obliterum forge now unlocked with a single quest (Firing up the Forge - 41778)


Ludovicus:
Thanks for the heads up.
All of the class guides needed an update and it was committed at https://github.com/Ludovicus/WoW-Pro-Guides/commit/bdabc0f0dd82396a0d4917cb2fed4662adb49a7f

PR #447, opened 2018-09-22
Update WoWPro_Broker.lua

elidion:
Added ILVL tag


Ludovicus:
the change to the parser is missing

PR #463, opened 2018-10-27
First try to restore Carbonite - WoWPro_Mapping relationship :)

Kaboca:
I checked with the Storheim guide - only works with the correct zone. Issue: Multiple map-points not removed until completion. Every kill restore them :)
You know WowPro better than I do.
So I suggest you check the results before the source code is merged!

### DIFF
--- a/WoWPro/WoWPro_Mapping.lua
+++ b/WoWPro/WoWPro_Mapping.lua
@@ -396,13 +396,17 @@ function WoWPro:MapPoint(row)
 		i = WoWPro.NextStepNotSticky(WoWPro.ActiveStep)
 	end
 
-
 	-- Removing old map point --
-	if LastMapPoint and LastMapPoint == i and #cache > 0 and cache[1].index == i then
+	if LastMapPoint and LastMapPoint == i and #cache > 0 and cache[i] and cache[1].index == i then
 	    WoWPro:print("MapPoint: LastMapPoint=%d [%.2f,%.2f@%d] in %s. No update needed.", LastMapPoint, cache[1].x, cache[1].y, cache[1].map, cache[1].zone)
 	    return
 	else
-        WoWPro:dbp("MapPoint: LastMapPoint=%s, #cache=%d, cache[1].index=%s, i=%d", tostring(LastMapPoint),  #cache, tostring(#cache > 0 and cache[1].index), i)
+		-- Because cached Carbonite mappoint is only a number - currently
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_Mapping.lua
```diff
@@ -396,13 +396,17 @@ function WoWPro:MapPoint(row)
 		i = WoWPro.NextStepNotSticky(WoWPro.ActiveStep)
 	end
 
-
 	-- Removing old map point --
-	if LastMapPoint and LastMapPoint == i and #cache > 0 and cache[1].index == i then
+	if LastMapPoint and LastMapPoint == i and #cache > 0 and cache[i] and cache[1].index == i then
```
Ludovicus: Good catch!

WoWPro/WoWPro_Mapping.lua
```diff
@@ -573,8 +579,8 @@ function WoWPro:MapPoint(row)
 			-- autoarrival == 0 is a no-op
 		end
 		TomTom.db.profile.persistence.cleardistance = OldCleardistance
-	elseif TomTom then
-		WoWPro:print("WoWPro:MapPoint2(%s@%s/%s)",coords,tostring(zone),tostring(zm))
+	elseif Nx then
+		WoWPro:print("Carbonite WoWPro:MapPoint2(%s@%s/%s)",coords,tostring(zone),tostring(zm))
```
Ludovicus: I think that a better approach might be to forsake the TomTom compatibility layer and just call the Nx native interface.   Carbonite seems to not be emulating the new TomTom interface, and emulating the old interfaces is rather horrid, having tried to write the code myself for TomTom.  So I'd rather just go the Nx functions and not have another layer of confusion.

Ludovicus: `Nx:TTSetTarget (mid, zx, zy, opts["title"], opts["callbacks"])`

Ludovicus: Not forgotten.  Hope to merge my version in the next few weeks.
I apologize for the delay.

Ludovicus-Maior: Closing pull request.   Functionality implemented in https://github.com/Ludovicus-Maior/WoW-Pro-Guides/commit/d8717fe264ac3e003999b7029a69fd4a4f953cda#diff-267bc39b782785b0267ae52b40a4220b

PR #484, opened 2019-08-25
update 31-40 Jame Horde Guide

Emmaleah:
reasonably final copy I think


Emmaleah:
this copy of the horde 31-40 is the one you want
the alliance 40-50 just deletes the "." in |Z|Stormwind City.| x2.  I could swear I had commited that before, but it claims that i havent...

PR #530, opened 2019-09-03
Update 15_18_Redridge.lua

elidion:
Overhaul of Guide.


Emmaleah:
Need the F step at the 3nd to be H

PR #623, opened 2019-09-13
Update 30_41_Jame_Chapter_I.lua

elidion:
Start of guide vetting process.
* Updated almost every coord, some were quite off.
* Changed some ACTIVE and AVAILABLE tags to make guide reloads tolerable.
* Added proper punctuation.
* Reordered some steps to make logical sense with least running.
* Lonebrow Journal had wrong itemID in L step.
* Feralas misspelled.
* Holy Crap the Thalanaar waypoint took you a mile out into the ocean in the upper NW corner, was off by 6,327 yards.
* Highperch Wyverns waypoint was way off.
* First Hearth back to Theramore just had a note to "Turn in quests". Added proper T steps and waypoints.


elidion:
I am only to line 45, still a ton more to do...though I did add those T steps down after the first hearth step.

PR #624, opened 2019-09-13
Update 12_20_Darkshore.lua

Hendo72:
Improved Druid quest line with notes, PRE tags, and substituted ACTIVE|####| for C tags. Step won't work/show if you haven't done the PRE or the quest isn't ACTIVE.

### DIFF
--- a/WoWPro_Leveling/Classic/Alliance/12_20_Darkshore.lua
+++ b/WoWPro_Leveling/Classic/Alliance/12_20_Darkshore.lua
@@ -78,8 +78,8 @@ F Rut'theran Village|QID|730|M|36.34,45.58|N|Take a ride to the village.|
 R Darnassus|QID|730|M|55.88,89.55|Z|Teldrassil|
 A Trouble in Darkshore?|QID|730|M|31.25,84.45|Z|Darnassus|N|Go to The Temple of the Moon except don't go inside. Take the right path around the building and you'll run into a dwarf.|
 A Lessons Anew|QID|6121|C|Druid|M|35,8|Z|Darnassus|N|From Mathrengyl Bearwalker on the second floor. Teleport to Moonglade after accepting.|
-T Lessons Anew|QID|6121|C|Druid|M|56.2, 30.5|Z|Moonglade|N|To Dendrite Starblaze.|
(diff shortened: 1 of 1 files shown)


WoWPro_Leveling/Classic/Alliance/12_20_Darkshore.lua
```diff
@@ -78,8 +78,8 @@ F Rut'theran Village|QID|730|M|36.34,45.58|N|Take a ride to the village.|
 R Darnassus|QID|730|M|55.88,89.55|Z|Teldrassil|
 A Trouble in Darkshore?|QID|730|M|31.25,84.45|Z|Darnassus|N|Go to The Temple of the Moon except don't go inside. Take the right path around the building and you'll run into a dwarf.|
 A Lessons Anew|QID|6121|C|Druid|M|35,8|Z|Darnassus|N|From Mathrengyl Bearwalker on the second floor. Teleport to Moonglade after accepting.|
-T Lessons Anew|QID|6121|C|Druid|M|56.2, 30.5|Z|Moonglade|N|To Dendrite Starblaze.|
-A The Principal Source|QID|6122|C|Druid|N|From Dendrite Starblaze.|
+T Lessons Anew|ACTIVE|6121|M|56.2, 30.5|Z|Moonglade|N|To Dendrite Starblaze.|
```
Ludovicus-Maior: Why switch to ACTIVE from QID||C|Druid|?   
With C|Druid| the quest is filtered out at the guide parsing stage and never shows even in the current guide.

elidion: Yeah I second this. I would close this pull request.

C and T steps don't show if you don't have the quest, you only need to use |C|Druid|, PRE, ACTIVE, AVAILABLE and such on the A steps (can have them on other steps like K, N, F, R, l ect. but talking about this use case.)

Redo this by just adding the PRE on the A step...At the end of the line please, not in the middle. Trying to keep conditionals at the end after the note for readability.

STEP title |QID|questID|M|coords|Z|zone if needed|N|note|PRE|prereq QuestID not needed on C or T steps|?|Any other conditions, loot, lvl, prof, class, race, ect.|

PR #655, opened 2019-09-20
Update WoWPro_Parser.lua

elidion:
add an else make PLAYER ni if x and y don't exist. l so validateMapCoords doesn't throw up.


Ludovicus-Maior:
nil is the right thing

PR #674, opened 2019-09-26
01 12 elwynn patch

geridoc:
Small improvements for Northshire Valley.


geridoc:
> I'm confused by the last change. Why move the accept from immediately after the turnin? Now it reads: 1 ) turn in Milly Osworth 2) go to echo Ridge and kill some Kobolds 3) come back to Milly and pick up a new quest.

Whoops, you're right, it should be:
C Skirmish at Echo Ridge
T Milly Osworth
A Milly's Harvest

PR #768, opened 2019-11-01
Update 25_26_Wetlands.lua

Cagomei:
Removed Report to Captain Stoutfist for Human and Night elf as this quest requires a quest chain from Loch Modan.


elidion:
I fixed this with the proper PREs, the Grail PRE check pass undid the way I did it orginally. Plus this shouldn't be removed for humans anyway.

PR #774, opened 2019-11-10
Update 20_25_Bitsem_Ashenvale.lua

Cagomei:
Vol'jin is dead, Eitrigg is accept his quest turn ins.


Ludovicus-Maior:
Sorry for the delay.

PR #790, opened 2019-12-11
Update 110_120_Emmaleah_Voldun.lua

Cagomei:
Missing zone tag(s)


Ludovicus-Maior:
Changed made to http://wow-pro.com/wiki/vol039dun_horde

PR #794, opened 2020-01-18
Update 120_120_Elidion_BlackEmpire.lua

elidion:
Looks like a quest was removed from PTR, commenting out just in case it makes a return. Also put in the improved defensive system order, not sure if this was removed or Ludo had an older  version of my PTR playthroughs.


luis-ortiz:
Cool!  I was wondering why I could not get that quest.  I was thinking it was because I had not finished the end game storyline.   The version you have me was back when we were debugging TomTom.  It had alot of notes and comments that I incorporated in haste.

Ludovicus-Maior:
Cool! I was wondering why I could not get that quest. I was thinking it was because I had not finished the end game storyline. The version you have me was back when we were debugging TomTom. It had alot of notes and comments that I incorporated in haste.

PR #819, opened 2020-04-04
Update 21_30_Hendo_HordeChapter1.lua

Hendo72:
Added a note to remind player that upon turning in Assault on Fenris Isle (QID 442)to destroy Bethor's Potion; as it is not needed anymore.
- The note will only show if the quest is neither available nor active.
- May need verification on proper syntax for a quest that has been turned in.


Hendo72:
Looking for a better way to do this.

PR #949, opened 2020-07-29
Rewrite guide list

Gethe:
Complete rewrite of the guide list.

- Uses Blizzard templates and UI functions so we aren't reinventing the wheel
- Centralized handling of tooltips, tabs, and frame creation
- Redone UI
- Tested on Beta, Live, and Classic

In the future, the new code can also be used and adapted for Current Guide and the actual guide window.

WoWScrnShot_072920_111214


Gethe:
To be clear, this does not require any changes to the guides themselves.

PR #957, opened 2020-08-20
Update 110_110_Blanc_Suramar.lua

Blanckaert:
fixed what Grail screamed about.  will also replace website page so changes dont get written over


Ludovicus-Maior:
Closes #882

PR #964, opened 2020-08-28
Update 20_25_Twists_Duskwood.lua

elidion:
Changed some of the guide flow to make more sense, after running 3 alts through duskwood and getting irritated with some direction, I decided to update the guide a bit.


Ludovicus-Maior:
Uploaded to web site.

PR #965, opened 2020-09-01
Shadowlands Initial Push

elidion:
* Stripped out LVL tags
* changed the level of everything in  guide to be 10-50 with a few exceptions while it gets worked out.
* Merged the 2 DH guides - left the vault one to see if there is a way to transfer the history with it, but removed it from guides.xml
* Touched up DK Intro Guide
* Renamed every file in the Neutral Folder - Want to make sure this does what Ludo wants before I made changes to Alliance/Horde.
*Probabaly some other stuff i'm forgetting, cleaned up a lot of other crap.


WoWPro_Leveling/Alliance/01_05_Bitsem_Shadowglen.lua
```diff
@@ -1,290 +1,41 @@
-
--- WoWPro Guides by "The WoW-Pro Community" are licensed under a Creative Commons Attribution-NonCommercial-NoDerivs 3.0 Unported License.
```
Ludovicus-Maior, line 2: The comments should not be removed.
They will get added back in.

PR #988, opened 2020-09-08
Major GuideList Overhaul

elidion:
**WoWPro\WowPro_Leveling.lua**

- Added additional allied races to WoWPro.Leveling.AlliedStartGuides.
- Added Exiles Reach Detection.
- Adjusted Newcharacter detection for DK/DH/Allied Races to appropriate new starter levels.

**WoWPro_Leveling\WoWPro_Leveling_GuideList.lua**
* Added LevelRefresh function to dynamically assign level ranges. (Original plan, didn't really need its own function, can probably be integrated into GetGuides() now that everything is in place.)
* GetGuides modified to use the level refresh function and check sortlevel.
* levelsort function seemed bugged, was comparing a.level to b.level and getting nil errors instead of a.guide.level and b.guide.level (I think before my changes, it wasn't really being called.)
* Integrated WoWPro.GuidelistReset boolean into Leveling:GetGuideListInfo() to work with the GOSSIP/CHROMIE_TIME CLOSE events.

**WoWPro\WoWPro.lua**
* Removed WoWPro:NewGuideLevels
* Updated WoWPro:GuideLevels to behave like the old NewGuideLevels did, This is now an override from the default dynamic level range assignments
* Created WoWPro:GuideSort for a sort override when needed. (Does not require GuideLevels, but if used in conjunction with GuideLevels it needs to come afterward.)
* Removed the QuestColor changes that were added yesterday, new level refresh handles the level scaling properly now.

**WoWPro\WoWPro_Broker.lua**
* Had to call WoWPro.Leveling:GetGuideListInfo() during WoWPro.LoadGuideReal(), it was checking the startlevel, now that GuideLevels is optional, this could be Nil until the list was viewed.

**WoWPro\WoWPro_Events.lua**
* Added event handler for CHROMIE_TIME_CLOSE
* WoWPro.GuidelistReset = true; for GOSSIP/CHROMIE_TIME CLOSE events.

**WoWPro_Leveling\{faction}\Every_Single_Guide.lua** (except classic stuff)
* Removed all WoWPro:NewGuideLevels
* Removed most WoWPro:GuideLevels, added them on guides that needed
* Added WoWPro:GuideSort to pretty much everything. Even though it isn't completely necessary, with the dynamic adjusting of guides, high levels and low levels could see slightly different orders, this standardizes it. Another thing of note with GuideSort is it's still bound to its expansion, so sort order is independant of other expansions. Basically every guide that we want to appear first for an expansion will always have a GuideSort of 1.


Ludovicus-Maior:
Nice!

PR #989, opened 2020-09-09
Update in combat

elidion:
Got the addon to function while in combat. The secret sauce was in setting the button frames to the InsecureActionButtonTemplate which allows it to appear but not be usable until combat ends. Only reason the wowpro frames were secured to begin with was because the buttons used SecureActionButtonTemplate and all 15 frames have an invisible placeholder of that button tied to it in case it needs to be used. That change alone was 90% of it.

There is obviously the 1 caveat of not being able to get or clear the bindings because  ClearOverrideBindings and GetBindingKey is protected. Just put a combat check on those, when combat ends the automatic guide update that happens will properly set these.

I didn't delete any of the WoWProDB.global.RecklessCombat or WoWPro.MaybeCombatLockdown() checks, and warnings, though I did comment out the warning, I actually just enabled reckless by default. So technically if this tests good we could clean that out. But just in case we need to revert I wanted to leave it intact. 

Guide behaves exactly as normal, no weird text issues or anything, if you are in combat and you get a new quest that pops up with an item or a target, you will still see everything that you normally would, the target/item buttons will show but just not be clickable until combat ends. This is the way I always dreamed of the addon working.

* Added a CLASSIC check to the chromie time event

* fixed a typo in the parser, when you reached the end of the guide with no nextguide it would say "Fini" instead of "Finished"

 * corrected a couple files that accedentially had double unscores in their name.


Ludovicus-Maior:
I will merge.  Will test on **Never Trust a Big Barb and a Smile** just to be sure.

PR #991, opened 2020-09-10
Update 01_12_Elwynn.lua

Hendo72:
Corrected A step to t step.


Hendo72:
This is already corrected in the first pull request

PR #998, opened 2020-09-12
Faction and SHADOWLANDS changes

elidion:
Changed all faction calls to only happen once and then set it to variable

Got rid of SHADOWLANDS, this is a faux retail build in its own branch that doesn't function in retail anyway. Got rid of the seperations of BFA content.


WoWPro/WoWPro_CompatUtil.lua
```diff
@@ -147,21 +147,21 @@ function WoWPro.QuestLog_GetInfo(questLogIndex)
...
-    if WoWPro.SHADOWLANDS then
+    if not WoWPro.CLASSIC then
         return _G.C_QuestLog.IsComplete(questID)
     else
         return select(6, _G.GetQuestLogTitle(_G.GetQuestLogIndexByID(questID)))
     end
 end
 function WoWPro.QuestLog_IsQuestFlaggedCompleted(questID)
-    if WoWPro.RETAIL then
+    if not WoWPro.CLASSIC then
```
Gethe, line 164: Any reason this and others were changed to `not WoWPro.CLASSIC` instead of `WoWPro.RETAIL`? They are logically identical, but I think the later is more readable.

elidion: Honestly I never really thought about it, for me it's opposite I guess lol, I feel it more readable this way.
I guess because I view RETAIL as the default code,  then CLASSIC is what is patched in to work in exceptions. When I want to search all the files for classic exceptions I can simply search WoWPro.CLASSIC and see it all. 

It honestly didn't even occur to me that RETAIL was even there, like I said I assumed it default, when I got to the end I saw that one straggler so I changed it for consistancy.

PR #1010, opened 2020-09-14
Hunter-run thru

Emmaleah:


Emmaleah:
closed - merged in paladin changes also in next request.

PR #1012, opened 2020-09-14
Update 01_12_Elwynn.lua

Hendo72:
Added some quests from Kharanos to ensure player is level 12 when they face Vagash. (I'm figuring 30-45 minutes depending upon luck with drops)


Ludovicus-Maior:
Vagash was fun as a priest at  level 10.

PR #1025, opened 2020-09-18
Bastion and Maldraxxus guides

elidion:
added Maldraxxus guide.

Finished main story review on bastion, still got about 120 more lines of sidestory to go over.


WoWPro_Leveling/Neutral/SL_Bastion.lua
```diff
@@ -1,12 +1,11 @@
 local guide = WoWPro:RegisterGuide('CagBastion', 'Leveling', "Bastion", 'Cagomei', 'Neutral')
```
Ludovicus-Maior, line 1: You should add in the copyright notice and we should check in a copy of it in GitHub

PR #1085, opened 2020-10-04
Update SL_Maw_Intro.lua

Emmaleah:


elidion (commented):
We need to stop uncommenting these out. They don't belong in the guide right now. Beta has this enabled but it won't be here for first time users. We still need to code in proper ToF detection and then add it here. We don't want this accidently going live.

PR #1105, opened 2020-10-08
Update SL_Bastion.lua

Cagomei:


Emmaleah:
I have incorporated these changes into my PR. There will be other differences as I moved where the H and h steps are, as well as a few quest chains to incorporate the TOF path.
Emmaleah

PR #1146, opened 2020-10-16
Chromie Alert Top of all guides.

elidion:
Added Chromie Time not on top of every guide under level 50 except for BFA guides.


Ludovicus-Maior:
Brute force has its place!

PR #1183, opened 2020-10-22
WIP:LW BFA

Blanckaert:
maybe one of these times I will actually get all the corrections in, in ONE pull.


Blanckaert:
old Leatherworking.lua needs to be removed, since name change to Leatherworking_BFA.lua

Ludo, I hope I finally got this right.

PR #1266, opened 2020-11-09
Patch 55

Cagomei:


Ludovicus-Maior:
The correct old Dalaran references are:
The Underbelly@Dalaran!Eastern Kingdoms
Dalaran City@Dalaran!Eastern Kingdoms

PR #1340, opened 2020-11-27
Add coordinates to Kala's Den entrance

rpotor:
Added a coordinate to the entrance of Kala's Den for the Bastion quest Tough Love, because I'm blind and I hate it when I have to look for cave entrances. :-)


Ludovicus-Maior:
Changed it up slightly.

PR #1398, opened 2020-12-09
Update 31_40_Hendo_HordeChapter2.lua

Hendo72:
Silly mistakes... QIDs don't work if they are the wrong ones.


Hendo72:
I'm going to add this change to my new branch instead,

PR #1438, opened 2020-12-19
Patch 47

Emmaleah:
changes per @Kriptoker WOD leveling experience


Emmaleah:
I note that only some of these guides have been converted to use $/RARE someone...should ... fix that :)

PR #1521, opened 2021-01-28
Update CATA_Ashenvale.lua

Hendo72:
- revamped Never Again! quest steps.
- added R step to Satynaar quests to eliminate CC
- U|45710| does not seem to work.

### DIFF
--- a/WoWPro_Leveling/Horde/CATA_Ashenvale.lua
+++ b/WoWPro_Leveling/Horde/CATA_Ashenvale.lua
@@ -88,23 +88,27 @@ A Rain of Destruction|QID|13798|M|72.20,57.65|N|From Durak.\n[color=FF0000]NOTE:
 R Exit the mine|QID|13798|M|73.31,59.81|QO|1|N|Head back outside the mine.|IZ|Splintertree Mine|
 C Rain of Destruction|QID|13798|M|74.19,62.95|QO|1;2|N|Climb one of the two guard towers and use the Accursed Ore to target Elves and Ancients until complete.|U|45598|
 T Rain of Destruction|QID|13798|M|72.20,57.65|N|To Durak.|
+R Exit the mine|QID|13841|M|73.31,59.81|N|Head back outside the mine.|IZ|Splintertree Mine|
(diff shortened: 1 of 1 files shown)


WoWPro_Leveling/Horde/CATA_Ashenvale.lua
```diff
@@ -88,23 +88,27 @@ A Rain of Destruction|QID|13798|M|72.20,57.65|N|From Durak.\n[color=FF0000]NOTE:
...
 T Rain of Destruction|QID|13798|M|72.20,57.65|N|To Durak.|
+R Exit the mine|QID|13841|M|73.31,59.81|N|Head back outside the mine.|IZ|Splintertree Mine|
 A All Apologies|QID|13841|M|73.33,62.13|N|From Splintertree Demolisher.|PRE|13798|RANK|1|
-K Gorgannon|QID|26449|M|84.08,71.07;84.24,77.20;87.41,78.98;89.57,76.78|CS|QO|0|N|Follow the waypoints to Gorgannon in Demonfall Canyon. Kill him, pick up the loot.|
-C Never Again!|QID|26449|M|82.72,77.91;81.38,78.12;81.12,79.32;80.94,80.05;79.53,80.65;78.41,81.57;78.38,83.84|CS|N|Follow the waypoints through the Canyon and up another path to Diathorus. Kill him, pick up the loot.|
-H Splintertree Post|QID|26449|N|[coords]If Hearthstone isn't ready, just run back to Splintertree Post.|
-T Never Again!|QID|26449|M|73.20,60.08|N|To Valusha.|
+R Demon Fall Canyon|QID|26449|M|84.21,72.25|N|Follow the road east out of Splintertree through Felfire Hill. Go south at the fork before the 2nd bridge.|
+K Gorgannon|QID|26449|M|87.41,78.98;89.57,76.78|CS|QO|1|N|Follow the waypoints to Gorgannon to kill him and loot the Sword.|
```
Ludovicus-Maior, line 94: K Steps that use a QO should use ACTIVE instead of QID.  
If this was a rank 2 quest and you had rank 1 set, it would show.
This will work as is, but is not best practice.

PR #1564, opened 2021-03-04
(no coordinates) bug fix in WoWPro_Broker.lua

bidaumm:
fixed text for (No coordinates) showing when user has show coordinates turned off in config 
also changed formatting of (No coordinates) to not start on newline due to adding two lines when no note is present.

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -965,9 +965,10 @@ function WoWPro:RowUpdate(offset)
             if zone then
                 note = note .. "@" ..zone
             end
-        elseif not coord and action and not WoWPro.Guides[GID].NoCoordsOK then
+        elseif WoWProDB.profile.showcoords and not coord and action then
             -- No coordinates, let them know!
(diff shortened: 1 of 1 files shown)


bidaumm:
accidently removed pull requests trying to clean up personal branches.

elidion (changes requested):
profile.showcoords is  if someone has the option to show coords or not.

nocoordsok is to tell the guide no coords is okay, like in a prof guide or something and you don't need to see the "No Coordinates" message because it isn't applicable, or possibly on a note.

We also want the new line when there is a note because it just looks tacky when it splits, and isn't consistent, new line means it is always in the same spot under the note.

Also the intention of "No Coordinates" originally was to let people know in the note if there were no coords and wondering if their arrow was broken. We got lots of reports of "Arrow doesn't work on this quest" because we had no coords there intentionally. This was meant to be independent of showcoords option which puts the coords at the end of the note.

In regards to the extra line when there is no note, maybe we can add an "if not note then" in there.

PR #1590, opened 2021-03-29
Update SL_Covenant.lua

Blanckaert:
step needed a QO to clear (shush dont tell wife I was editing while playing WOW with her.)


Ludovicus-Maior:
Bad @Blanckaert !

PR #1671, opened 2021-05-24
Make zones recognizable on non-EN-clients

Arby69:


Arby69:
have to figure out why the lib "LibBabble-Subzone..." folder didn't make it to the pull request...

PR #1701, opened 2021-06-02
Update INTRO_BCC_Blood_Elf.lua

Hendo72:
Removed the note at the end and changed `GuideNextGuide` to CLASSIC_BC_Ghostlands as it seems to be intended.

I accidentally pushed this change to my Main and it was too late to undo.


Hendo72:
I'm going to undo my changes and do this properly.

PR #1776, opened 2021-07-23
Update VAN_BCC_10_19_Ghostlands.lua

Hendo72:
I forgot to include the T step for `The Warsong Reports|QID|6543|` in Ghostlands.


Ludovicus-Maior:
😭

PR #1855, opened 2021-11-05
Update INTRO_BCC_DwarfGnome.lua

Hendo72:
Coordinates pointing in wrong direction


Hendo72:
Cagomei will be taking care of these changes.

PR #1894, opened 2021-12-24
Updated Argent Tournament guide for Horde

nzall:
Updated most of the Argent Tournament guide (WowPro_Dailies/Horde/WotLK_argTour_H.lua). This isn't entirely finished yet, and not everything is tested yet. I pretty much need someone to double check my work, because after 4 hours of looking up Wowhead quests, they tend to blur together.

things to verify:

- faction-specific coordinates for Valiant and Champion quest targets may still point towards the Alliance location;
- Not all Crusader quests are implemented yet, only those of 2021-12-24 (Deathspeaker Kharos and The Fate of the Fallen).
- Quest accept and turnin coordinates are "close enough" locations that are within a few feet in front of the group of NPCs associated with the quests, but not 100% matching the exact NPC locations;
- Quest IDs and NPC names for Valiant dailies might be wrong. I tried to directly convert the existing Alliance code in the guide 1-1 to Horde counterparts, but there were only 3 races done in the existing guide, so Troll and Tauren had to be written manually. I based the accepts of their dailies on Undead, but turnin on Blood Elf (which in hindsight might have been a confusing choice). Some of the quest IDs may additionally be missing from completion step QID lists. I was interrupted halfway through updating them and may have missed some;
- Quests that have racial requirements may be missing race listings from Cataclysm and later expansions;
- In general, this guide can benefit from additional testing by characters who aren't crusaders yet.

### DIFF
--- a/WoWPro_Dailies/Horde/WotLK_ArgTour_H.lua
+++ b/WoWPro_Dailies/Horde/WotLK_ArgTour_H.lua
@@ -199,21 +199,36 @@ A Among the Champions|QID|13814|PRE|13795|C|Death Knight|M|73.59,20.09|N|From Il
 A Taking Battle To The Enemy|QID|13813|PRE|13795|C|Death Knight|M|73.80,19.45|N|From Zor'be the Bloodletter.|FACTION|Horde|
 
 ; Crusader Quests
-A Deathspeaker Kharos|QID|14105|M|69.96,23.44|CC|N|From High Crusader Adelard|
-A The Fate of the Fallen|QID|14107|M|69.96,23.44|CC|N|From Crusader Silverdawn|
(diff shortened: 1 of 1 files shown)


Ludovicus-Maior:
Looks like a good start.  Merging the different factions will not be that hard afterwards.

nzall:
@Ludovicus-Maior You should be aware that there are still some minor issues left that I'm aware of: some of the Troll and Tauren Valiant dailies have wrong QIDs (using the same QID for the same quest for different races); and the new Races that were added in Cataclysm and later (Goblin, Pandaren and Legion/BfA allied races) probably aren't flagged correctly in some quests. Is it okay if I make an issue for those or comment on an existing issue?

PR #1943, opened 2022-03-06
9.1.5 changes

Hendo72:
Reference and Innuendo changes that were made in patch 9.1.5
[Patch 9.1.5 (undocumented changes)](https://wowpedia.fandom.com/wiki/Patch_9.1.5_(undocumented_changes))
I believe I've gotten all of them. I hope I haven't broken anything by making these changes untested.

I also added the missing apostrophe to 'Combatants'


Blanckaert (commented):
The combatants changes in the profession guide when in the [item=] [spell=] will crash.   Special characters (ie ') not allowed in markup.

PR #1966, opened 2022-03-24
Update VAN_BCC_10_19_Ghostlands.lua

Hendo72:
- added Level 22 check


luis-ortiz:
Failed check is something I need to fix.

PR #1981, opened 2022-04-19
Update Guides.xml

Cagomei:
I very stupidly forgot to update the Guide file when I updated the guide name.


elidion:
This was already taken care of in my pull request.

PR #2063, opened 2022-08-06
horde wrath changes

Cagomei:
Copies over BC to Wrath

Changed 2 to 3

Added DK starter guide.

fixed guide file.


Ludovicus-Maior:
@Cagomei Conflicts!

PR #2071, opened 2022-08-11
Slime quest corrections -  Winterspring

Hendo72:
- Added the missing QID and updated |ACTIVE|
- trying to remember why I was using the '-' in the ACTIVE. I removed it for now.


Ludovicus-Maior:
There were merge errors, but it was simple to clean them up.

PR #2116, opened 2022-09-22
Update WoWPro_Profession_Parser.lua

Spoony36:
Slightly Reword profession parser as i think it would look better in the profession guides

image

Reject if needed and you'd prefer to leave as is

### DIFF
--- a/WoWPro_Profession/WoWPro_Profession_Parser.lua
+++ b/WoWPro_Profession/WoWPro_Profession_Parser.lua
@@ -71,7 +71,7 @@ function WoWPro.Profession:PreRowUpdate(row)
 								CraftQty = 0
 							end
 							WoWPro.note[k] = ""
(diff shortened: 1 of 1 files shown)


Blanckaert:
Yeah, reason I had "crafting" was a hope that we could get it to automate.  But that is something that is not being figured out (any time soon[tm])

Spoony36:
Oh so it would auto update that text like 
crafting up to 60
crafting up to 59
crafting up to 58 etc?

That would be a nice feature tbh

Ludovicus-Maior:
Oh, I think I had a patch to do that that I never finished debugging...
Sigh.

Spoony36:
I'm just wondering if the wording needs tweaking then

image

Should it maybe say 

Craft these from Cooking level 1 to 40 (You may need to craft up 60 to achieve this)

Ludovicus-Maior:
Discuss with @Blanckaert .  When you both agree, a pull request can be submitted.
I rarely get involved with wording issues as English is not my first language.

PR #2118, opened 2022-09-22
Basic Cooking Guide

Spoony36:


Blanckaert (commented):
On the M Steps it should be [item=#/item name] not [spell=] and you don't need |SPELL| on M Lines, it's |ITEM|

PR #2122, opened 2022-09-25
Alliance tador Guide: Remove broken pre-requiste

mdavidson37:
Removing a prerequiste that prevented picking up the starting quest in Talador.


Cagomei (commented):
From my research this correction is good. that A step is telling you to pick up 34560 OR 34561, not sure why it would have a PRE for those two quests.

PR #2131, opened 2022-10-01
Azure Span Recorder

Spoony36:
Will Need edits and tidy up


Spoony36:
Needs some TLC from pro authors :)

PR #2136, opened 2022-10-03
Update WotLK_Engineering.lua

EzeRage:
Fixes the Goblin / Gnomish detection in the guide.


Blanckaert (commented):
This was testing for needed in later steps,  I'll look though it later today.  But according to the one that put the charges in,  it works now.

PR #2176, opened 2022-10-16
Update DF_Waking_Shores.lua

Emmaleah:
started thru with an alliance char.


elidion:
Newer version was posted

PR #2179, opened 2022-10-17
Update DF_Waking_Shores.lua

Emmaleah:
I actually changed almost every line, in that I changed the Z to be 1022;The Waking Shores instead of just 1022. testing it out. I was wishing it would make the text in game show the zone name, but it does not, so then I deleted all of that...  fixed some errors from my bad cut and paste last night and got thru to where I was recording the one area not yet done... and wow crashed during that. so lost it all (   ... save early ... save often ...


elidion:
There is a new better version.

PR #2250, opened 2022-11-11
Update DF_Waking_Shores.lua

Spoony36:
Pre removed


Spoony36:
Cagomei, re-running guide and will correct the issues

PR #2283, opened 2022-11-24
Update WOTLK_VAN_30_40_HordeChapter2.lua

EzeRage:
Removed quest A New ore Sample because i spent over 40 minutes killing mobs for quest drop (there are a total of 8 spawns in the entire zone) and didnt get a single drop. So i removed the quest due to bad RNG possibilites and time vs exp gained isnt much of a loss for the player. Quest turn in is worth the same as farming kills for 10 minutes with unrested and no heirloom gear.  Also changed run to quest turn in to Hearthstone. (There should be no reason the players hearth is on cooldown as i was killing fast and using a shortcuts to get around (Slow Fall and Blink) and my hearth was never on cooldown when i was supposed to run to turn in quests.


Cagomei (changes requested):
Need to remove one of the | on line 106

EzeRage:
I will have to correct this tomorrow

EzeRage:
Fixed :)

PR #2357, opened 2022-12-18
Update to WinterVeil.lua for Horde

OftKilted:
Updated level requirement for free metzen to align with WowHead level guidance of level 30, wording for 'Hillsbrad Foothills' and there were no snow mounds around cave. updated to suggest manually closing step if no snow mounds.


Ludovicus-Maior:
It is considered polite to make the same corrections to both guides if they are faction specific.
Also, the LVL requirements are needed only for the first quest.

PR #2367, opened 2022-12-22
Update DF_Thaldraszus.lua

EzeRage:
Ugh hopefully this "fixes" it.


Cagomei:
We need to actually do some research and figure out what the triggers are instead of just commenting them out. They are required quests to unlock specific world quests. If we need to adjust the DFREN tags or whatever so be it. But commenting them out isnt accomplishing anything.

PR #2430, opened 2023-03-08
C_Container. correction WotLK

Hendo72:
Updated trash step with new variable


Ludovicus-Maior:
Oh, in  WRATH?!?!?  
They were supposed to leave that alone!

PR #2573, opened 2023-12-05
Update WoWPro_Config.lua

Spoony36:
Attempt at new config setup though i have no coding knowledge so this might need someone to shit on it :D

Currently loads with no errors but does it work? 

image

Things to do

- Needs looking at the overall code to see if this is viable
- Guide Display & Current guide needs looking at as i dont know how to incorporate that yet
- Attempt needed to fix the right click of mini-map button opening guide list
- Right click of header to open guide list directly needs fixing
- Probably lots more :)

Do not merge until someone with more knowledge than me, well mainly copilot who has guided me in this... has checked it all

### DIFF
--- a/WoWPro/WoWPro_Config.lua
+++ b/WoWPro/WoWPro_Config.lua
@@ -1,13 +1,9 @@
--- luacheck: globals pairs ipairs
-
------------------------------
---      WoWPro_Config      --
------------------------------
-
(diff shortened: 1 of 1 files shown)


Spoony36:
This is not the version we are going with, Tabbed Version is the way

PR #2590, opened 2023-12-23
Fixes the question Mark Bug in Levelling for Allaince for Menu Display

Spoony36:
Covenant campaign is proving to be stubborn


Ludovicus-Maior:
This has been fixed by https://github.com/Ludovicus-Maior/WoW-Pro-Guides/commit/a340f3b5ff67b1f9705359be12265410bc0f74d8 instead.

PR #2610, opened 2023-12-30
Wrong minimum Level

capau:
Just got my DH to level 60 and i can do these quests.


Ludovicus-Maior:
On hold till a check on Conditions is verified.

PR #2614, opened 2024-01-03
1/3/24

Cagomei:
corrections for placement on the guide list.


Ludovicus-Maior:
Oh, that fixes the Dalaran message that I had not tracked down yet too!

PR #2657, opened 2024-02-06
Update LunarFestival.lua

Hendo72:
Travel steps still need coordinates to be updated. Documented where required.


Ludovicus-Maior:
I assume this is a placeholder due to the **{update Coords}** notes in there.

PR #2658, opened 2024-02-06
Direct Opening of Tabs Fix

Spoony36:
:D

Just helping relieve the pressure on current fixes needed


Spoony36 (commented):
Current Guide Config can be discarded as Ace doesn’t support a scrollable frame.

dropdown menu should be good though

PR #2691, opened 2024-03-15
Update WoWPro_Broker.lua

Spoony36:
Remove Carbonite as i think Tom Tom is the addon of choice nowadays?

Check everything is correct and merge or reject if you wish to keep Carbonite as an option


Ludovicus-Maior:
Not now.  Have to play nice.

PR #2705, opened 2024-03-30
Update WoWPro_Broker.lua

Spoony36:
Hide the Carbonite/TomTom timer from printing in chat upon login


Ludovicus-Maior:
If you see the countdown time at all, something is wrong.
It does not print normally at all on my client.
I'm going to reject this, as it is a valid warning.

PR #2707, opened 2024-03-31
Frame Resize to fit buttons in advanced mode

Spoony36:
Fix frame size when in advanced mode

BEFORE
image

AFTER
image


Spoony36:
Need to review this so closing

PR #2710, opened 2024-04-01
Hopefully Resolve Current Guide and GuideList Issues

Spoony36:
Check and if Happy then Merge!


Spoony36:
redoing as widgets so its the correct way

PR #2717, opened 2024-04-15
Frame settings between versions.

Spoony36:


Spoony36:
Like this @Ludovicus-Maior Ludo

Ludovicus-Maior:
@Spoony36 We are now quite content!

PR #2762, opened 2024-05-24
Cata Header correction

Hendo72:
All of the Cata Daily guides have been updated to fix "WoWPro/WoWPro.lua:1199: attempt to index local 'guide' (a string value)" error.

"Cata" needs to be added as the last variable (It's normally '4')

The Achieve, Profession, and World Event guides will need this done as well. The headers are not modernized with the correct License.


Hendo72:
I am closing this PR and submitting a new one that works correctly.

PR #2766, opened 2024-05-25
WoWPro_Trade.lua TLC

Spoony36:
Add Cata to WoWPro_Trade.lua


Ludovicus-Maior:
This will not quite work, but it is a good start @Spoony36

Ludovicus-Maior:
In particular, this should break retail.

PR #2771, opened 2024-05-26
Cata 3 Menu System Workaround

Spoony36:
go from this 

image

to this
image


Ludovicus-Maior:
Naw.   Having "Profession"/"Dailies" at the top level is correct.
The next level should just be a guide list.

PR #2779, opened 2024-05-31
Migrate EasyMenu

mdavidson37:
- Created new compatibility function for EasyMenu (which was removed in the latest Alpha build breaking the new style guide selection)
- Migrate the  Context Menus to the new function
- Pass to migrate any calls to removed API functions I missed last time onto the new compatibility functions
- Update the new Compatibility functions to be backwards compatible with Retail.

**Known Issue**

The latest Alpha build has broken the TomTom Addon so the integration isn't currently working.


Ludovicus-Maior:
@mdavidson37 Please post the TomTom bug you see here and I'll see if I can take care of it.
I am the secondary maintainer of the addon!

PR #2789, opened 2024-06-15
Handle Deprecation of C_Scenario.GetCriteriaInfo

mdavidson37:
Blizzard have moved GetCriteriaInfo into C_ScenarioInfo.

The new function does not return an equivalent to quantitiyString - but that value isn't used by anything in the addon so I'm OK with removing it.


Ludovicus-Maior:
Thanks!  Just what was needed!

PR #2794, opened 2024-06-23
'basic' localisation

Spoony36:


Spoony36:
What other Languages would be Needed?

PR #2797, opened 2024-06-29
Rename all authors to WoWPro Team in Alliance and Neutral

Spoony36:


Spoony36:
Realised this would mess with next guide stuff

PR #2804, opened 2024-07-08
Handle deprecation of GetItemIcon

mdavidson37:
Looks like I missed this case in the first pass.

Bonus update - Updated the hearthstone list with the new additions since 9.2.0.


Ludovicus-Maior:
Excellent

PR #2864, opened 2024-09-08
Prevent Error When Menulist is Nil

Spoony36:
image


Ludovicus-Maior:
@Spoony36 This will result is no menu being presented.   The question is, why is menuList nil?
The failure occurred before.

Spoony36:
Ah I thought it would just stop the error from showing but I’ll close then.

PR #2882, opened 2024-09-30
Upcoming Blizzard Deprecations

Furydins:
Did a sweep of the list on Warcraft Wiki of pending Blizzard API deprecations (just 2 thankfully) and migrated over to the new functions - fingers crossed this will keep us ahead for a bit!

I've switched to checking the API for existence of the functions (instead of game version) as I suspect we'll be seeing these move to classic at some point.


Ludovicus-Maior:
@Furydins Excellent!

PR #2919, opened 2024-11-18
Track Progress bar %age for use with QO steps

Furydins:
This will allow the |QO|1<10| tags to use the progress bar % in the same way as a normal x/y step.


Ludovicus-Maior:
Well, that was simpler than I expected!

PR #2947, opened 2025-01-03
Disable Addons for LUA error debugging

Spoony36:
add new command /wp disable-addons which will disable all addons apart from WoWPro and WoWPro_Levelling, BugSack,BugGrabber, TomTom to establish if addons are tainting.

Add a second command /wp enable-addons for user to re-enable all addons that were disabled from running the above command.

One small issue that i haven't been able to figure out and that is how to disable Dependencies like this 

image

### DIFF
--- a/WoWPro/WoWPro_Command.lua
+++ b/WoWPro/WoWPro_Command.lua
@@ -122,6 +122,44 @@ local function handler(msg, editbox)
             end
         end
         _G.ChatFrame1:AddMessage("Blizzard API stored in: <World of Warcraft>/WTF/Account/<#>/SavedVariables/WoWPro.lua")
+    elseif ltoken == "disable-addons" then
+        local keepEnabled = {
+            "WoWPro",
+            "WoWPro_Leveling",
+            "BugSack",
+            "!BugGrabber",
+            "TomTom",
+        }
+    
+        WoWProCharDB.disabledAddons = {}
+    
+        local function isAddonKept(name)
+            for _, addon in ipairs(keepEnabled) do
+                if name == addon then
+                    return true
+                end
+            end
+            return false
+        end
+    
+        for i = 1, C_AddOns.GetNumAddOns() do
+            local name, _, _, enabled = C_AddOns.GetAddOnInfo(i)
+    
+            if enabled and not isAddonKept(name) then
+                C_AddOns.DisableAddOn(name)
+                table.insert(WoWProCharDB.disabledAddons, name)
+            end
+        end
+    
+        ReloadUI()
+    
+        for _, name in ipairs(WoWProCharDB.disabledAddons) do
+            C_AddOns.EnableAddOn(name)
+        end
+    
+        WoWProCharDB.disabledAddons = {}
+        ReloadUI()
+    
     else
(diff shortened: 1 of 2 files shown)


Furydins:
The dependencies won't load if their dependency is missing - so I doubt that really matters.

WoWPro/WoWPro_Command.lua
```diff
@@ -122,6 +122,40 @@ local function handler(msg, editbox)
...
+
+        for i = 1, _G.C_AddOns.GetNumAddOns() do
+            local name, _, _, enabled = _G.C_AddOns.GetAddOnInfo(i)
+            if enabled and not isAddonKept(name) then
+                _G.C_AddOns.DisableAddOn(name)
+                table.insert(WoWProCharDB.disabledAddons, name)
+            end
+        end
+        _G.ReloadUI()
+
```
Furydins, line 153: Am I being dense or is the else clause for enable-addons missing?

Spoony36: I’ll have to check later when home but am sure it works as is

Spoony36: @Ludovicus-Maior - Thoughts on this?

Ludovicus-Maior: Echoing what @Furydins said, after you do `/wp disable-addons`, how do you undo what it did?

Spoony36: but i'll look at separating the commands and will revert

Spoony36: this approach better?

```    elseif ltoken == "disable-addons" then
        local keepEnabled = {
            "WoWPro",
            "WoWPro_Leveling",
            "BugSack",
            "!BugGrabber",
            "TomTom",
        }
    
        WoWProCharDB.disabledAddons = {}
    
        local function isAddonKept(name)
            for _, addon in ipairs(keepEnabled) do
                if name == addon then
                    return true
                end
            end
            return false
        end
    
        for i = 1, C_AddOns.GetNumAddOns() do
            local name, _, _, enabled = C_AddOns.GetAddOnInfo(i)
    
            if enabled and not isAddonKept(name) then
                C_AddOns.DisableAddOn(name)
                table.insert(WoWProCharDB.disabledAddons, name)
            end
        end
        ReloadUI()
    
    elseif ltoken == "enable-addons" then
        if WoWProCharDB.disabledAddons then
            for _, name in ipairs(WoWProCharDB.disabledAddons) do
                C_AddOns.EnableAddOn(name)
            end
            WoWProCharDB.disabledAddons = {}
            ReloadUI()
        end```

Furydins: That looks like it should do the trick!

Ludovicus-Maior: @Spoony36 Yeah, go with this and we can merge!

Spoony36: Done, Merge if all ok

Spoony36:
@Ludovicus-Maior If happy ready to merge i think

PR #2951, opened 2025-01-08
Removed a double word

Justin-1993:
found a double word while using the guide, went in and removed it


Spoony36:
@Justin-1993 Thanks for the PR

PR #2960, opened 2025-02-03
Quest handin NPC update

Furydins:


manuelaidos123:
Thanks for the update. Could you confirm if this change affects only The Skittering Breach quest or if Valrin Paller was removed from other quests as well? A screenshot of the new NPC location would be helpful for verification.

Furydins:
Valrin was still the handin for the Mycomancer Cavern delve quest. 

Wowhead has also detected the the change for Skittering Breach: https://www.wowhead.com/quest=83768/delvers-call-the-skittering-breach

PR #2981, opened 2025-02-25
Seasonal purge is only supported by classic

Furydins:


WoWPro/WoWPro.lua
```diff
@@ -573,7 +573,7 @@ function WoWPro:OnEnable()
         WoWPro:dbp("PlayerGetTimerunningSeasonID function is not available")
     end
 
-    if _G.C_Seasons.GetActiveSeason() then
+    if WoWPro.CLASSIC and _G.C_Seasons.GetActiveSeason() then
```
Ludovicus-Maior, line 576: How about:
`if  _G.C_Seasons and _G.C_Seasons.GetActiveSeason() then`
instead?

PR #3002, opened 2025-03-28
Zone changes for cata

Cagomei:


Cagomei:
IDK how much this breaks anything. The cata zone data has not been updated since it was added back in beta.

PR #3015, opened 2025-04-05
Guide percentage

Spoony36:
Please check i'm not missing anything, this is to change the values from 100/249 completed to a Percentage on main Frame TitleBar

image

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -1443,6 +1443,8 @@ if step then
             if note ~= "" then
                 if action == "B" then
                     note = "Buy " .. nomen .. " " .. note
+                elseif action == "M" then
+                    note = "Craft " .. nomen .. " " .. note
                 else
                     note = "Kill and loot " .. note
                 end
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_Broker.lua
```diff
@@ -1443,8 +1443,6 @@ if step then
             if note ~= "" then
                 if action == "B" then
                     note = "Buy " .. nomen .. " " .. note
-                elseif action == "M" then
```
Furydins, line 1446: Did you mean to remove this?

Furydins (commented):
Did we mean to remove support for M steps?

WoWPro/WoWPro_Broker.lua
Spoony36: How did that get there!

Spoony36: resolved and added back in.

Spoony36:
I need to find out why it isn't replicated across all clients though i've just noticed

Cata

image

Spoony36:
We still have an issue if the user skips several steps as the percentage only shows for instance 6% even though end of guide has been reached because technically 6% is actually completed.


image

PR #3016, opened 2025-04-05
Update WoWPro_Broker.lua

Spoony36:
Change Mainframe to show percentage rather than steps.

Please check implementation is correct before committing

image


Furydins (approved):
Nice!

PR #3017, opened 2025-04-05
Bugfixes for Auto show/hide

Furydins:
**Bugs Fixed**

- Changing zone when inside an instance would prevent the guide from showing after leaving the instance.
- Changing zone when in combat could cause the guide to show prematurely.

In order to cope with overlapping hide events the previous logic used a tracking counters, however the number of  hide and show events could in the above cases be unbalanced causing the counter to lose track of the true state.

To address this I've reworked the logic to instead check the character and guide state on every event to determine if the guide should be hidden or shown. This should be more reliable and will self-heal in the event of any issues.


WoWPro/WoWPro_Parser.lua
```diff
@@ -489,7 +489,7 @@ end
 function WoWPro.EmitStep(i)
     local GID = WoWProDB.char.currentguide
 
-    if not i then return "<nil>"; end
+    if not i or not WoWPro.stepcount then return "<nil>"; end
```
Furydins, line 492: Prevent the parser blowing up if an auto show event occurs when no guide is loaded.

PR #3048, opened 2025-05-11
% progress now ignores sticky steps

Furydins:
The %age progress is now based upon the main active step istead of the first sticky step which will avoid it appearing "stuck" when we have long lived sticky steps.

Also deleted some unused code and duplicated steps.


WoWPro/WoWPro_Broker.lua
```diff
@@ -1723,29 +1723,17 @@ function WoWPro.UpdateGuideReal(From)
     end
     WoWProCharDB.Guide[GID].progress = p
     WoWProCharDB.Guide[GID].total = WoWPro.stepcount - WoWPro.stickycount - WoWPro.optionalcount
-    -- Display progress based on the configuration
```
Furydins, line 1726: No need to set title here as we set it again a few lines below!

PR #3059, opened 2025-05-31
Update BC_Hellfire.lua

Spoony36:
@Cagomei - Can you check these edits and make sure they are correct?


Spoony36:
@Cagomei is going to do as she's a pro and i'm not!

PR #3078, opened 2025-07-02
Cata close out

Cagomei:
Archived all the cata files.

Updated MoP Zone data.


Cagomei:
I know its a huge PR. I moved ALL of the cata files from the main folders to the archive. Take as long as you need to push this. if there are problems. lemme know.

PR #3082, opened 2025-07-07
Begin Modularizing Broker Possibly?

Spoony36:
Thoughts on reducing broker and moving some code into their own files or leave well alone :D 

@Ludovicus-Maior 
@Furydins


Furydins:
As it looks fine to me I'm gunna do the developer thing and nit pick!

How about calling WoWPro_BrokerUtils just WoWPro_Utils

My reasoning is to decouple the functions from the idea they are just for the broker.

PR #3083, opened 2025-07-08
New menu system attempt, will need expert eyes!

Spoony36:
Currently set to draft but IF EVERYONE thinks this is ok and correct, then good to be merged.

Retail Seems to be ok

image
image

Classic - Seems to be ok (Falls back to old system)

image

MOP - Seems to be ok (Looks like it uses new Menu?)

image

PTR 11.2 - No errors

image


Spoony36:
> @Spoony36 You got a few conflicts with main. Do you want to tackle them or shall I try?

This may need revisiting with all the recent changes, i'll try and take a look soon

PR #3085, opened 2025-07-13
Title bar fix

Spoony36:
if you double click titlebar to collapse it, then move your character a little distance, the frame expands again which we don't want.

This adds a persistence state.


Spoony36:
closing as i didn't want to update wowpro.lua

PR #3091, opened 2025-07-21
trying fo fix missing modules.

Cagomei:
I give up i cant get it to work.


WoWPro_Dailies/Classic_MoP/WoWPro_Dailies_Cata.toc
```diff
@@ -0,0 +1,17 @@
+## Interface: 40402
```
Furydins, line 1: Needs to be renamed to WowPro_Dalies_Mists.toc, updated to mops version no and moved up to the root of WoWPro_Dalies!

WoWPro_Profession/WoWPro_Profession_MoP.toc
```diff
@@ -0,0 +1,15 @@
+## Interface: 40402
```
Furydins, line 1: Needs to be renamed to WoWPro_Profession_Mists.toc and updated for the mists client version

Furydins (changes requested):

WoWPro_Dailies/Classic_MoP/WoWPro_Dailies_Cata.toc
Furydins: Also need to update the last 3 lines to point to the MoP directory (currently still cata)

PR #3094, opened 2025-07-23
Faction Selection Bugfix

Furydins:
WoW-Pro permanently stores the faction when starting up and does not currently update it when a neutral character selects their faction.

This change will reset the faction in the NEUTRAL_FACTION_SELECT_RESULT even handler and should (hopefully) get the guide to behave itself when players leave the Pandren starting area.


Ludovicus-Maior:
Grrrrrr.!

PR #3102, opened 2025-08-06
Update WoWPro_Parser.lua

Spoony36:
Sample code for experts to look at and tweak and implement if needed.

If not required please delete/close


WoWPro/WoWPro_Parser.lua
```diff
@@ -1,4 +1,4 @@
--- luacheck: globals tostring tonumber
+-- luacheck: globals tostring tonumber time GetCurrentRegion WoWPro_IsValidDate
```
Furydins, line 1: As you are using _G to access GetCurrentRegion - you shouldn't need to list it here.

WoWPro/WoWPro_Parser.lua
```diff
@@ -1285,3 +1328,37 @@ function WoWPro.SetupGuideReal()
     end
     WoWPro:SendMessage("WoWPro_PostLoadGuide")
 end
+
+-- Validate date components to ensure they form a valid date
+function WoWPro_IsValidDate(year, month, day, hour, min)
```
Furydins, line 1333: I suggest changing this to WoWPro.IsValidDate - this should allow you to remove it from the luacheck header above!

PR #3111, opened 2025-08-22
Possible Fix for random Quest adding due to WoWPro.inhibit_oldQuests_update = false

Spoony36:
@Ludovicus-Maior - Isn't this what you suspected of adding the random quest?

If so i think this will fix it?


Ludovicus-Maior:
hmmm, it’s pretty dammed close.  let me play with it!

PR #3123, opened 2025-09-06
Move Current Guide and Guidelist out of Blizz options - Starting point for dev's to tweak/rip apart/edit :D

Spoony36:
This moves both frames out of Blizz options and opens them as standalone frames to prevent issues with frame placement on large monitors/resolutions as per discord tickets

image

This will make mainframe have tab like buttons for those 2 windows to open/close 

image

image

image

Also slightly tweaked/Modernized Guide list to use more modern tabs

Blizz Options will now only have true config options

image

TO DO: 
Choose appropriate icons

Future Improvements if others agree:
Add reset guide button
Add skip steps button
Add Discord Link Button

### DIFF
--- a/WoWPro/WoWPro_Config.lua
+++ b/WoWPro/WoWPro_Config.lua
@@ -1190,23 +1190,6 @@ local function createActionConfig()
     return actions
 end
 
-local function createGuideConfig()
-    local actions = {
-        name = L["Guide Selection"],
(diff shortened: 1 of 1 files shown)


Furydins (commented):
I like that we're separating out core functionality like guide selection from the game options page!

Furydins (approved):

PR #3131, opened 2025-09-13
Further Frame Tweaks

Spoony36:


WoWPro/WoWPro_Frames.lua
```diff
@@ -1220,8 +1220,11 @@ function WoWPro:CreateNextGuideDialog()
 end
 
 
+local isResettingGuide = false
 function WoWPro.ResetCurrentGuide()
-    if not WoWProDB.char.currentguide then return end
+    if isResettingGuide then return end
```
Ludovicus-Maior, line 1228: A simpler change would be:
```
function WoWPro.ResetCurrentGuide()
    if not WoWProDB.char.currentguide then return end
    if not WoWPro.GuideLoaded  then return end
    WoWPro.GuideLoaded = false
     …
 ```
 and use `WoWPro.GuideLoaded` as the guard.
 The use of the timer and flag will not be reliable.

Ludovicus-Maior (changes requested):

PR #3201, opened 2025-11-29
warrior class hall

Cagomei:


Ludovicus-Maior:
… a little more than just the warrior class hall!

PR #3258, opened 2026-02-15
"DONE" label in progress column in guide list

Hendo72:
**Add localized "DONE" label to guide list when guide reaches completion**

Updated guide list completion display to use a localized DONE label and to respect the current guide's title-bar completion state, so a guide that reaches 100% in the title bar shows DONE even when the completion-based ratio is still below 1. Added a helper that mirrors title-bar logic and wired it into all guide list modules. Also added localized strings for DONE across supported locales.

**In its current state, if all of the steps are completed (100% is shown in titlebar), it will display DONE. However, as with any of the guides, the guide must be loaded at least once to get the current progress.

### Files touched:
- WoWPro.lua
- WoWPro_Localization.lua
- WoWPro_Leveling_GuideList.lua
- WoWPro_Achievements_GuideList.lua
- WoWPro_Dailies_GuideList.lua
- WoWPro_Profession_GuideList.lua
- WoWPro_WorldEvents_GuideList.lua

**Draft PR suggestion (for all guides): persist the title-bar percent per guide (e.g., WoWProCharDB.Guide[GID].percent) during guide updates and consume that cached percent in the guide list to mark DONE for any guide, not just the current one. This is a small change in the update loop where title-bar percent is computed, plus a list-side check against the stored percent.

### DIFF
--- a/WoWPro/WoWPro.lua
+++ b/WoWPro/WoWPro.lua
@@ -1281,6 +1281,29 @@ function WoWPro:GetGuideProgress(guideID)
     return
 end
 
+function WoWPro:IsCurrentGuideTitlebarComplete(guideID)
+    local currentGuide = WoWProDB.char.currentguide
+    if not currentGuide then
+        return false
+    end
+    if guideID and guideID ~= currentGuide then
+        return false
+    end
+    local total = WoWPro.stepcount or 0
+    if total <= 0 then
+        return false
+    end
+    if WoWProDB.profile.guideprogress then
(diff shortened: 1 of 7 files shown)


Hendo72:
image
If you want to see what it looks like. 
Westfall is actually done, but it has 2 surviving steps that never "complete" and therefor will never be 100%

Ludovicus-Maior:
@Hendo72 There have been some major changes since this was written.  
I think it needs to be closed and redone, picking and choosing the real code changes amongst the stuff that was moved for the refactoring/modularization.

PR #3298, opened 2026-03-30
Enforce AutoCompleteVerify on all step completions

Hendo72:
### TL;DR
- Added `WoWPro.AutoCompleteVerify(step)` as required validation for any completion.
- Added enforcement in `WoWPro:CheckFunction` and `WoWPro.CompleteStep` so completion only proceeds if verify passes.
- Added `UpdateGuideBusy`/`UpdateGuidePending` guard logic to avoid overlapping update calls (prevents timing race completion).

### If you want the details:

1. **Added global “completion safe check” logic**
   - Added `WoWPro.AutoCompleteVerify(step)` in WoWPro_Broker.lua.
   - _Purpose_: single source of truth to decide if a step may be completed.
   - _Logic_:
        - no QID → true
        - quest in current log and complete → true
        - quest historically complete (`IsQuestFlaggedCompleted`) → true
        - for `C` steps with objective text: verify each objective via `QuestObjectiveStatus`
        - otherwise → false
    - _Why_: to prevent unstable or premature completion (true “are you sure” verification).
2. **Added mandatory enforcement in check-event path**
   - _Updated_ `WoWPro:CheckFunction(row, button, down)` in WoWPro_Broker.lua:
        - on right-click completion attempt, call `WoWPro.AutoCompleteVerify(row.index)`
        - if verify returns false, cancel and don’t complete
   - _Why_: prevent forced check-box completion from bypassing validation.
3. **Added mandatory enforcement in generic step completion**
   - Updated `WoWPro.CompleteStep(step, why, noUpdate)` in WoWPro_Broker.lua:
        - if not `noUpdate`, call `WoWPro.AutoCompleteVerify(step)`
        - if false, abort completion and return false
   - _Why_: ensures all completion methods (auto-engine or code path) are gated by the same check.
4. **Working in tandem with race-prevention flags**
   - `UpdateGuideBusy` / `UpdateGuidePending` guards exist already
   - They prevent re-entrancy during `UpdateGuide` and support the safety check
5. **Practical behavior effect**
   - New state: step completion happens only if verified by `AutoCompleteVerify`.
   - If verification fails, completion does not happen, and guide stays consistent.
   - No external option controls this; it’s an unconditional safeguard.
6. **Confirmation in final implementation**
   - Added inspectable check points and debug logs:
        - `WoWPro:dbp("AutoCompleteVerify failed ...")` in both paths.
   - _Why_: for maintainability and clear debugging for future testers.

`With assistance from Raptor mini AI engine`

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -18,6 +18,10 @@ WoWPro.mygroupsteps = {}
 WoWPro.myGroupTrack = {}
 WoWPro.playerGroup = {}
 
+-- Update concurrency guards (prevent race conditions from nested / overlapping update loops)
+WoWPro.UpdateGuideBusy = false
+WoWPro.UpdateGuidePending = false
+
 -- Debug toggles
 WoWPro.DEBUG_STICKY_PAIRING = false -- Set to true to enable sticky pairing debug output
 WoWPro.DEBUG_REPEATABLE = false -- Set to true to enable debug output for repeatable A step resets and quest log changes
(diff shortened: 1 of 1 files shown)


Hendo72:
I'm going to close this branch and focus on the bug that actually came out of this rabbit hole.

PR #3302, opened 2026-04-02
Move Autocomplete Logic to it's own file

Spoony36:
As before, reject if you'd prefer to leave as is


Ludovicus-Maior:
Well, at least  there are a few .TOC files there !

Spoony36:
Realised that this is probably not the best to do at this stage

PR #3303, opened 2026-04-03
Remove Author Column from guidelist and add search box

Spoony36:
image

image

Requested a while back

### DIFF
--- a/WoWPro/WoWPro_GuideList.lua
+++ b/WoWPro/WoWPro_GuideList.lua
@@ -37,6 +37,98 @@ do -- GuideListMixin:SetTooltip
         self.module:SetTooltip(guide)
     end
 end
+
+local function NormalizeSearchText(text)
+    text = tostring(text or ""):lower()
+    return text:match("^%s*(.-)%s*$")
+end
+
+local function GuideFieldMatches(value, filterText)
+    if value == nil then
+        return false
+    end
+    if type(value) == "function" then
+        value = value()
+    end
+    return tostring(value):lower():find(filterText, 1, true) ~= nil
+end
+
+local function GetVisibleHeaderInfo(headerInfo)
+    if not headerInfo or not headerInfo.names then
+        return headerInfo
+    end
+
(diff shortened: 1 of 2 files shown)


Ludovicus-Maior:
OOOOOOh!

WoWPro/WoWPro_GuideList.lua
```diff
@@ -37,6 +37,98 @@ do -- GuideListMixin:SetTooltip
...
+            return true
+        end
+    end
+
+    local guideData = guide.guide or {}
+    return GuideFieldMatches(guide.GID, filterText)
+        or GuideFieldMatches(guide.Name, filterText)
+        or GuideFieldMatches(guide.Zone, filterText)
+        or GuideFieldMatches(guide.Content, filterText)
+        or GuideFieldMatches(guide.Author, filterText)
```
Ludovicus-Maior: If Author is not displayed, why search on it?

WoWPro/WoWPro_Templates.lua
```diff
@@ -71,6 +71,7 @@ function ScrollListMixin:RefreshLayout()
     local guides = self.guides
     local buttons = _G.HybridScrollFrame_GetButtons(self)
     local offset = _G.HybridScrollFrame_GetOffset(self)
+    local headerCount = self.activeHeaderCount or #self.headers
```
Ludovicus-Maior, line 74: Will Author be on the header list and thus the count off by one?

Ludovicus-Maior (changes requested):
Two questions

WoWPro/WoWPro_GuideList.lua
Spoony36: Opps, i wasn't intending to remove the author column when i added the search function :D

WoWPro/WoWPro_Templates.lua
Spoony36: It shouldn't be but have simplified the search function

PR #3308, opened 2026-04-06
Fix zone auto-complete bug during guide load

Hendo72:
### What the bug is
- During guide load, a zone-based step can be auto-completed by AutoCompleteZone()
- But the guide was not fully refreshed afterward
- That left the step marked complete in data, while the active guide UI still showed it as the current active step

**In short:** AutoCompleteZone() completed a F/H/b/P/R step during setup, but the guide load process did not re-synchronize the guide state and the completed step would still appear active in the guide window.

### TL;DR
- Fix targets the load-time edge case where zone auto-completion and guide refresh were out of sync.
- It makes AutoCompleteZone() settle during setup, then refreshes once if needed.
- It also adds safety fallback for pathological cases and avoids extra updates on normal zone changes.

### What changed in WoWPro_Parser.lua
Added a bounded retry loop around WoWPro.AutoCompleteZone() inside SetupGuideReal().
- The code now:
    1. sets WoWPro.GuideLoaded = "Loaded"
    2. runs WoWPro:AutoCompleteQuestUpdate(nil) and initial WoWPro:UpdateGuide("WoWPro.SetupGuideReal(1)")
repeatedly calls WoWPro.AutoCompleteZone() while it keeps completing steps
    3. stops when no more zone auto-completions occur
- Safety controls were added (as per Furydin's suggestion):
    - maxZoneAttempts limit
    - maxZoneTime timeout
    - If those limits are exceeded, the code now aborts the loop and reloads the guide cleanly instead of continuing indefinitely.

### What changed in WoWPro_Events.lua
- The ZONE_CHANGED event handler was tightened so WoWPro:UpdateGuide(event) only runs when WoWPro.AutoCompleteZone() actually returns true.
- That means the guide is refreshed only when a zone-based step was completed, not on every zone change.

### Why this fixes it
- The parser change makes sure load-time zone auto-completions are detected and the guide is refreshed after they happen.
- The event change avoids unnecessary updates and keeps ZONE_CHANGED behavior aligned with actual state change.
- The loop is bounded so addon initialization cannot hang or risk Blizzard killing the addon because of excessive load-time work.

** each file change was done in its own commit for clarity.

### DIFF
--- a/WoWPro/WoWPro_Events.lua
+++ b/WoWPro/WoWPro_Events.lua
@@ -572,7 +572,7 @@ WoWPro.RegisterEventHandler("ZONE_CHANGED", function(event, ...)
         WoWPro.AutoHideFrame("|cff33ff33Instance Auto Hide|r: " .. event, "INSTANCE")
     end
     if WoWPro.Ready(event) then
-        if WoWPro.AutoCompleteZone(...) then
+        if WoWPro.AutoCompleteZone() then
             WoWPro:UpdateGuide(event)
         end
     end
--- a/WoWPro/WoWPro_Events.lua
+++ b/WoWPro/WoWPro_Events.lua
@@ -5,7 +5,6 @@
 --  WoWPro_Events.lua   --
 --------------------------
 
-local L =  WoWPro_Locale
 local successfulRequest = _G.C_ChatInfo.RegisterAddonMessagePrefix("WoWPro")
 -- Are we ready to roll?
 function WoWPro.Ready(who)
@@ -312,12 +311,14 @@ end
 
 
 -- Auto-Complete: Quest Update --
-function WoWPro:AutoCompleteQuestUpdate(questComplete)
+function WoWPro:AutoCompleteQuestUpdate(questComplete, newQuests, missingQuests)
     local GID = WoWProDB.char.currentguide
     if not GID or not WoWPro.Guides[GID] then return end
     if not WoWProCharDB.Guide then return end
     if not WoWProCharDB.Guide[GID] then return end
     if not WoWProCharDB.Guide[GID].completion then return end
+    newQuests = newQuests or {}
+    missingQuests = missingQuests or {}
 
     WoWPro:dbp("Running: AutoCompleteQuestUpdate(questComplete=%s)",tostring(questComplete))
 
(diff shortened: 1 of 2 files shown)


Hendo72:
I am confirming that this fix works. I have an active situation with an h step that was doing this exact thing and the fix is clearing the step and moving forward.

Ludovicus-Maior:
I'm concerned this is too complicated a solution and that we may have already implemented the logic to retry `WoWPro:UpdateGuide` properly during the guide load stage.

At the end of `WoWPro.SetupGuideReal()` there is the following code:
```
    WoWPro:UpdateGuide("WoWPro.SetupGuideReal(1)")
    -- Location, Location, Location
    if WoWPro.AutoCompleteZone() then
        WoWPro:UpdateGuide("WoWPro.SetupGuideReal(2)")
    end
```
It should just be:
```
    WoWPro:UpdateGuide("WoWPro.SetupGuideReal()")
```
In `function WoWPro.AutoCompleteZone()`:
Replace:
```
    if action == "F" or action == "H" or action == "b" or action == "P" or action == "R" then
        if not WoWProCharDB.Guide[WoWProDB.char.currentguide].completion[currentindex] then
            if (step == zonetext) or (step == subzonetext) then
                WoWPro.CompleteStep(currentindex,"AutoCompleteZone:"..step)
                return true
            end
            if (targetzone == zonetext) or (targetzone == subzonetext) then
                WoWPro.CompleteStep(currentindex,"AutoCompleteZone:"..targetzone)
                return true
            end
            local _, _, mapId = WoWPro:GetPlayerZonePosition()
            if (tonumber(targetzone) and tonumber(targetzone) == mapId) then
                WoWPro.CompleteStep(currentindex,"AutoCompleteZone:"..targetzone)
                return true
            end
        end
    end
    return false
```
With:
```
   if action == "F" or action == "H" or action == "b" or action == "P" or action == "R" then
        if not WoWProCharDB.Guide[WoWProDB.char.currentguide].completion[currentindex] then
            if (step == zonetext) or (step == subzonetext) then
                WoWPro.CompleteStep(currentindex,"AutoCompleteZone:"..step)
                WoWPro:UpdateGuide("Action "..action.." matches ".. zonetext.."/".. subzonetext)
                return
            end
            if (targetzone == zonetext) or (targetzone == subzonetext) then
                WoWPro.CompleteStep(currentindex,"AutoCompleteZone:"..targetzone)
                WoWPro:UpdateGuide("Action "..action.." matches ".. zonetext.."/".. subzonetext)
                return
            end
            local _, _, mapId = WoWPro:GetPlayerZonePosition()
            if (tonumber(targetzone) and tonumber(targetzone) == mapId) then
                WoWPro.CompleteStep(currentindex,"AutoCompleteZone:"..targetzone)
                WoWPro:UpdateGuide("Action "..action.." matches ".. tostring(mapId))
                return
            end
        end
    end
```

This will queue up an update of the guide at a future time and retry automatically until the guide loads.
It already has a retry timer.  Oh and remove any tests for a return value from `WoWPro.AutoCompleteZone()`.

Hendo72:
This has already been added to the code through alternate means.

PR #3310, opened 2026-04-10
Fix autocomplete questupdate

Hendo72:
### Bug explanation
1. A step false auto-complete
    - A step completion was being triggered and decided by WoWPro.newQuest alone.
    - That signal is only a “new quest seen” hint from the log scan, not definitive proof the accept step was actually matched and active.
    - As a result, an A step could auto-complete prematurely.

2. T step missingQuest scalar bug
    - WoWPro.missingQuest was a single retained QID.
    - that scalar could be overwritten or stale across quest-log changes.
    - as a result, turn-in completion could use the wrong missing QID or lose the correct one.

### TL;DR: 
- A step fix: require the quest to actually exist in WoWPro.QuestLog before auto-completing on newQuest
- T step fix: replace the stale single WoWPro.missingQuest scalar with QID-specific WoWPro.missingQuests
    - **Result**: turn-in completion now matches exact missing quest IDs and avoids stale/mismatched T auto-complete state

### What changed
1. **A step fix**
    - WoWPro.newQuest is still used as the initial match signal.
    - But accept auto-complete now also requires that WoWPro.QuestLog[QID] exists.
    - newQuest remains as a trigger, not the final decision.
2. **T step fix**
 - converted missingQuest from a scalar into QID-specific missing-quest tracking
 - PopulateQuestLog() now records WoWPro.missingQuests[QID] = true
 - AutoCompleteQuestUpdate() now only completes a T step when the exact missing QID matches
 - the matched missing QID is cleared immediately after use
### Why it works
- the A step fix prevents premature completion from newQuest alone
- the T step fix prevents stale scalar missingQuest state from mis-matching later turn-ins 
    - (I did ask a lot of questions as to why this was necessary)

_**Written with help/guidance from Raptor mini AI**_

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -4297,8 +4297,10 @@ function WoWPro.PopulateQuestLog()
     end
     local newQuests = {}
     local missingQuests = {}
-    WoWPro.newQuest = false
-    WoWPro.missingQuests = missingQuests
+    -- Legacy scalar quest state, preserved only for compatibility.
+    -- New code should use explicit diff tables instead.
+    -- WoWPro.newQuest = false
+    -- WoWPro.missingQuests = missingQuests
 
     -- Generating the Quest Log table --
     WoWPro.QuestLog = tablecopy(WoWPro.FauxQuestLog) -- Reinitiallizing the Quest Log table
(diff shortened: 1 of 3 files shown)


Hendo72:
I'm leaving this as draft for now.
This is the bug that I derived from the previous PR and filtered it down to this.

Ludovicus-Maior:
@Hendo72 I would consider a different approach.
The Broker is going to tests the quest status for any uncompleted quest for *X* steps.
I'm not sure we need WoWPro.newQuest or WoWPro.missingQuest at all!
Perhaps we can just delete that code!

Hendo72:
I think this code is ready to be used.

I will point out that I've avoided making changes to the Recorder code despite it still uses the legacy stuff. I don't want to break that anymore than it already is.

Ludovicus-Maior:
This is a good change that was broken due to the events refactoring.
I'll do it, @Hendo72 .

PR #3312, opened 2026-04-16
Fix sticky boundary logic for S!US steps

Hendo72:
### The Bug
- The sticky boundary was being calculated from `WoWPro.ActiveStep` by default.
- When the visible row block started earlier than the current active step (`k < WoWPro.ActiveStep`), the display logic still used `ActiveStep` as the cutoff.
- That meant sticky rows could be considered eligible too early, before the guide had actually reached the visible start row.
- The fix makes the boundary use `k` when `k` is earlier than `ActiveStep`, so sticky visibility is anchored to the current visible start instead of the later active step.

**The bug was not an exception in `S!US` handling.**
- It was in the sticky display filter inside `WoWPro:RowUpdate()`.
- When the visible row window started before `WoWPro.ActiveStep`, the code still used `ActiveStep` as the sticky cutoff.
- That made sticky rows appear too early, including `S!US` steps, because visibility was being allowed based on a later step index instead of the current visible start.

So the fix is about the boundary used to decide whether a sticky step is allowed to show, not about `S!US` parsing or the exception logic itself.

### Why only `S!US` was noticeable
- S!US is a special case: the first part is a sticky `S` step and the second part is an unsticky US step.
- The bug was in the sticky visibility filter for sticky steps, especially for `C`-type stickies with quest objectives.
- Normal `S` steps usually only show when the guide has reached them and they don’t have the paired `US` activation/completion handshake, so they did not trigger the visible bug as clearly.
- `S!US` steps are more sensitive because:
    - the `S` part is sticky,
    - the `US` part is hidden until that `S` part is completed,
    - and the boundary logic decided whether a sticky step was “after the current active step.”
- That made the `S` half of an `S!US` pair appear too early, while plain `S` steps stayed within normal sticky visibility rules.

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -1150,6 +1150,9 @@ function WoWPro:RowUpdate(offset)
     -- Now sort: stickies first, then regular
     local completion = WoWProCharDB.Guide[GID].completion
     local stickyBoundary = WoWPro.ActiveStep or k
+    if WoWPro.ActiveStep and k < WoWPro.ActiveStep then
+        stickyBoundary = k
+    end
     local stickySteps = {}
     local regularSteps = {}
     for _, stepIdx in ipairs(allSteps) do
@@ -1164,16 +1167,14 @@ function WoWPro:RowUpdate(offset)
                 local available = WoWPro.available and WoWPro.available[stepIdx]
                 local activeReq = WoWPro.active and WoWPro.active[stepIdx]
 
-                local isSUS = WoWPro.sticky[stepIdx] and WoWPro.unsticky[stepIdx]
-                -- Never show sticky steps that are beyond current progression (except S!US, which stays visible until its condition completes)
-                if not isSUS and stepIdx > stickyBoundary then
+                -- Never show sticky steps that are after the current active step
+                if stepIdx > stickyBoundary then
                     showSticky = false
 
                 -- Respect AVAILABLE/ACTIVE tags for sticky visibility (filters, not triggers)
-                -- S!US should always show until completion, regardless of these filters
-                elseif not isSUS and available and not WoWPro.QuestAvailable(available, false, "AVAILABLE") then
+                elseif available and not WoWPro.QuestAvailable(available, false, "AVAILABLE") then
                     showSticky = false
-                elseif not isSUS and activeReq and not WoWPro:QIDsInTableLogical(activeReq, WoWPro.QuestLog) then
+                elseif activeReq and not WoWPro:QIDsInTableLogical(activeReq, WoWPro.QuestLog) then
                     showSticky = false
                 -- For C steps with QID and QO, check if specific objective is incomplete
                 elseif action == "C" and QID and questtext then
@@ -1212,8 +1213,7 @@ function WoWPro:RowUpdate(offset)
                     end
                 -- For other sticky types, show if we've reached that step
                 else
-                    -- S!US steps should remain visible while incomplete, even if they are past the current active step
-                    if isSUS or stepIdx <= k then
+                    if stepIdx <= k then
                         showSticky = true
                     end
                 end
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -1149,6 +1149,8 @@ function WoWPro:RowUpdate(offset)
 
     -- Now sort: stickies first, then regular
     local completion = WoWProCharDB.Guide[GID].completion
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_Broker.lua
```diff
@@ -1149,6 +1149,8 @@ function WoWPro:RowUpdate(offset)
 
     -- Now sort: stickies first, then regular
     local completion = WoWProCharDB.Guide[GID].completion
+    -- Never show sticky steps that are after the current active step.
```
Ludovicus-Maior, line 1152: Does this prevent nested S steps ?
They are all over the place.

Ludovicus-Maior (changes requested):

WoWPro/WoWPro_Broker.lua
Hendo72: What are nested S steps? Are you refering to QIDs with multiple steps?

This did bring up a potential issue that I presented to the AI. What if the ActiveStep is zone-filtered (for example) and the next step is a S step? That S step is now visible and gets filtered to the bottom of the stickies (that's how the visual works). However, when you walk back into the zone, the hidden activestep now becomes visible and the S step will be hidden because it's not eligible to be displayed yet.

We have a potential fix for that, but I'm wondering if we should bother. I would assume that S step depends on the previous step being completed or they would've had it above it.

Ludovicus-Maior: Here is an example from Nagrand:
```
l Howling Wind|QID|9861|M|64.0,45.1|Z|107; Nagrand|N|Kill mobs until you get the Howling Wind.|L|24504|S!US|
C Windroc Mastery|QID|9854|M|64,36|Z|107; Nagrand|N|Kill the Windrocs needed for this quest.|S|
C Talbuk Mastery|QID|9857|M|61.00,43.00|Z|107; Nagrand|N|Kill the Talkbuks needed for this quest, they are around the trees.|S|
C Clefthoof Mastery|QID|9789|M|64.00,48.00|Z|107; Nagrand|N|Kill the Clefthoofs needed for this quest.|
A The Howling Wind|QID|9861|M|64.0,45.1|Z|107; Nagrand|N|Accept the quest from the item.|U|24504|O|
T The Howling Wind|QID|9861|M|60.5,22.4|Z|107; Nagrand|N|To Elementalist Morgh.|
A Murkblood Corrupters|QID|9862|PRE|9861|M|60.5,22.4|Z|107; Nagrand|N|From Elementalist Morgh.|
T The Throne of the Elements|QID|9869|M|60.5,22.4|Z|107; Nagrand|N|To Elementalist Sharvak.|
A A Rare Bean|QID|9800|M|60.5,22.4|Z|107; Nagrand|N|From Elementalist Lo'ap.|
A Muck Diving|QID|9815|M|60.5,22.4|Z|107; Nagrand|N|From Elementalist Lo'ap.|
A The Underneath|QID|9818|M|60.5,22.4|Z|107; Nagrand|N|From Elementalist Untrag.|
T The Underneath|QID|9818|M|61.8,24.4|Z|107; Nagrand|N|To Gordawg. He wanders around.|
A The Tortured Earth|QID|9819|PRE|9818|M|61.8,24.4|Z|107; Nagrand|N|From Gordawg.|
l I Must Have Them!|ACTIVE|10109|M|65.29,48.29|QO|1|Z|107; Nagrand|N|Kill Air Elementals and other mobs until you get 3x Air Elemental Gas.|S|
C The Tortured Earth|QID|9819|M|67.00,36.00|Z|107; Nagrand|N|Kill the Earth Spirits needed for this quest.|
C Windroc Mastery|QID|9854|M|64,36|Z|107; Nagrand|N|Kill the Windrocs needed for this quest.|US|
C Talbuk Mastery|QID|9857|M|61.00,43.00|Z|107; Nagrand|N|Kill the Talkbuks needed for this quest, they are around the trees.|US|
l I Must Have Them!|ACTIVE|10109|M|65.29,48.29|QO|1|Z|107; Nagrand|N|Kill Air Elementals until you get 3x Air Elemental Gas.|US|
```

Hendo72: The way the code is setup, if the step is an S or an S!US, it's filtered to the top of the window and will remain there as long it's valid with no visibility restrictions active. 

Now, I can't say for certain where the S step will be displayed in relation to the order at the top of the window. I'd like to believe that the new step is added to the bottom of the sticky pile.

Ludovicus-Maior:
@Hendo72 This one needs some TLC from you since I'm not 100% up to date on all the S/US changes.

PR #3318, opened 2026-04-22
Update WoWPro_NilGuide.lua

Spoony36:
Update no Guide Loaded, feel free to choose better terminology


Ludovicus-Maior:
Good catch!

PR #3321, opened 2026-04-24
Correct WoWPro.GuideLoaded and add WoWPro.GuideUpdated

Ludovicus-Maior:
Change WoWPro.GuideLoaded to be a pure boolean and factor out WoWPro.GuideUpdated instead of 
using a magic truthy value in WoWPro.GuideLoaded.

### DIFF
--- a/WoWPro/WoWPro.lua
+++ b/WoWPro/WoWPro.lua
@@ -440,6 +440,7 @@ function WoWPro:OnInitialize()
     WoWPro.DebugClasses = (WoWPro.DebugLevel > 0) and WoWProCharDB.DebugClasses
     WoWPro.GossipText = nil
     WoWPro.GuideLoaded = false
+    WoWPro.GuideUpdated = false
     -- Selector is Deprecated
     WoWProDB.profile.Selector = nil
     if type(WoWProDB.profile.checksoundfile) == "string" then
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -554,6 +554,7 @@ function WoWPro:LoadGuide(guideID)
         WoWProDB.char.currentguide = WoWPro:GuideFormalName(guideID)
     end
     WoWPro.GuideLoaded = false
+    WoWPro.GuideUpdated = false
     WoWPro.current_strategy = nil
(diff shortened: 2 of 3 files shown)


WoWPro/WoWPro.lua
```diff
@@ -440,6 +440,7 @@ function WoWPro:OnInitialize()
     WoWPro.DebugClasses = (WoWPro.DebugLevel > 0) and WoWProCharDB.DebugClasses
     WoWPro.GossipText = nil
     WoWPro.GuideLoaded = false
+    WoWPro.GuideUpdated = false
```
Furydins, line 443: Am I being blind or is WowPro.GuideUpdated never read by anything?

Does it even need to exist?

WoWPro/WoWPro_Broker.lua
```diff
@@ -2251,11 +2249,8 @@ function WoWPro.UpdateGuideReal(From)
         end
     end
     WoWPro:MapPoint()
+    WoWPro.GuideUpdated = true
     WoWPro:SendMessage("WoWPro_PostUpdateGuide")
-    if WoWPro.GuideLoaded ~= "Updated" then
-        WoWPro.ZONE_CHANGED_NEW_AREA("ZONE_CHANGED_NEW_AREA_GUIDE_LOAD")
```
Furydins, line 2256: I think this was the only thing that relied on Loaded and Updated status not being the same. 

Assuming the removal is intended(?) - this would mean GuideUpdated is indeed redundant!

WoWPro/WoWPro.lua
Ludovicus-Maior: Correct.  It _will_ be used by `WoWPro.Ready` as part of the punt/replay code.  It will be used in `WoWPro.Ready()` and to control the even punt and replay code.

WoWPro/WoWPro_Broker.lua
Ludovicus-Maior: Not used **yet**! In fact WoWPro.GuideLoaded is more danger as the changes progresses.  But I don't want to break existing code.

Furydins (approved):
I guess I passed the "was he awake test"

All good!

PR #3324, opened 2026-04-24
Fix some bugs

Spoony36:
This PR cleans up two low-risk issues in the zone utility code:

Removes an unreachable duplicate number branch from the table pretty-printer in WoWPro_Zones.lua

Fixes misleading duplicate-map debug messages so DefineZone3, DefineZone4, and DefineZone5 log their actual function names in WoWPro_Zones.lua

Why
The pretty-printer had a dead copy/paste branch that could never execute, which added noise and made the type handling harder to read.

The zone registration helpers for client versions 3, 4, and 5 were emitting DefineZone2 in their duplicate override debug output. That did not affect runtime behavior, but it made debugging duplicate zone registrations misleading.

Impact
No intended functional behavior change to map registration
Cleaner and more maintainable type handling in the debug table serializer
More accurate debug output when duplicate zone mappings are detected

### DIFF
--- a/WoWPro/WoWPro_Zones.lua
+++ b/WoWPro/WoWPro_Zones.lua
@@ -52,7 +52,7 @@ end
 function WoWPro.DefineZone3(mapId, zone, mapType, parent_map, group_id, ... )
     if WoWPro.Client ~= 3 then return end
     if WoWPro.Zone2MapID[zone] then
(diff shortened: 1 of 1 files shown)


Ludovicus-Maior:
LOL, copy/paste trash!

PR #3325, opened 2026-04-25
Recorder Fixes

Spoony36:
Fix four bugs in WoWPro_Recorder

- OnDisable() was unregistering events from the wrong frame (GuideFrame) using an empty local list; now correctly unregisters WoWPro.Recorder.events from WoWPro.RecorderFrame 
- Quest picker in Add Step dialog used tinsert(questList, QID, info.title) which silently fails for sparse/high quest IDs; changed to questList[QID] = info.title
-  "Useable Item ID" field getter fell back to returning stepInfo.optional instead of stepInfo.use (copy/paste error)
-  After registering a step via Add Step, questtextset was set to true instead of nil, blocking objective auto-prefill on subsequent adds

### DIFF
--- a/WoWPro_Recorder/WoWPro_Recorder.lua
+++ b/WoWPro_Recorder/WoWPro_Recorder.lua
@@ -55,9 +55,10 @@ end
 
 function WoWPro.Recorder:OnDisable()
     -- Unregistering Recorder Module Events --
-    local events = {}
-    for _, event in ipairs(events) do
-        WoWPro.GuideFrame:UnregisterEvent(event)
+    if WoWPro.RecorderFrame and WoWPro.Recorder.events then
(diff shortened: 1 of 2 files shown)


WoWPro_Recorder/WoWPro_Recorder.lua
```diff
@@ -55,9 +55,10 @@ end
 
 function WoWPro.Recorder:OnDisable()
     -- Unregistering Recorder Module Events --
-    local events = {}
-    for _, event in ipairs(events) do
-        WoWPro.GuideFrame:UnregisterEvent(event)
+    if WoWPro.RecorderFrame and WoWPro.Recorder.events then
```
Ludovicus-Maior, line 58: Oh so stupid!

PR #3328, opened 2026-04-25
Recorder Improvements (Potential?)

Spoony36:
Feel free to reject or change if needed

Subject:
recorder: default advanced mode and add edit-step jump navigation

Body:

make Advanced mode the default for new users
preserve existing users’ saved Advanced preference (true/false)
add Go To Step controls directly in Edit Step dialog
validate jump range and keep guide viewport/scrollbar synced to selected step

This means you can edit live guides directly in edit step now with a jump to step button :)

All live guides now show when using the open a guide to edit button

image

When using the go to step option, the step jumps to the step number you input for direct editing

image

Edited Step

Edited Step

Added a 'Current Guide' Button to the edit step screen so you can see the full guide and know which step you wish to work on

image

### DIFF
--- a/WoWPro_Recorder/WoWPro_Recorder.lua
+++ b/WoWPro_Recorder/WoWPro_Recorder.lua
@@ -29,12 +29,11 @@ function WoWPro.Recorder:OnInitialize()
 end
 
 function WoWPro.Recorder:OnEnable()
-    if WoWProCharDB then
-        if WoWProCharDB.Advanced == nil then
-            WoWProCharDB.Advanced = true
-        end
-        WoWPro.Recorder.Advanced = WoWProCharDB.Advanced
+    WoWProCharDB = WoWProCharDB or {}
+    if WoWProCharDB.Advanced == nil then
+        WoWProCharDB.Advanced = WoWPro.Recorder.Advanced
     end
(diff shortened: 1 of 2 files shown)


Ludovicus-Maior:
Hmm, some of your images are blank unless I click on them!

WoWPro_Recorder/WoWPro_Recorder.lua
```diff
@@ -29,9 +29,12 @@ function WoWPro.Recorder:OnInitialize()
 end
 
 function WoWPro.Recorder:OnEnable()
-	if WoWProCharDB then
-		WoWPro.Recorder.Advanced = WoWProCharDB.Advanced or false
-	end
+    if WoWProCharDB then
+        if WoWProCharDB.Advanced == nil then
+            WoWProCharDB.Advanced = true
```
Ludovicus-Maior: So what if I don't want advanced mode?
If that is the only mode, then some more changes may be appropriate.

Spoony36: Basic mode is still supported.

I changed initialization so we no longer hard-force Advanced to true on enable; we now read/write a persisted per-character preference and mirror it into runtime state. Toggle still switches modes and persists, so users who do not want Advanced can keep it off.

PR #3333, opened 2026-04-28
Add optional warband quest completion support for guide step filtering

Spoony36:
This change adds support for treating warband-completed quests as completed in the guide engine, while keeping existing character-only behavior as the default.

What changed:

Added a compatibility wrapper that checks for available account/warband quest completion APIs and safely falls back when unavailable. 
Updated quest completion caching logic to optionally merge character completion and warband completion.
Added a new profile setting to enable or disable warband quest completion checks. 
Reset relevant completion caches when the setting changes so results update immediately. 
Added a profile default for the new setting, defaulting to disabled for backward compatibility. 

Behavior:

Default behavior is unchanged (character-only completion). 
When enabled, quest completion checks return true if either: the current character completed the quest, or
another character in the warband completed it.
On clients where warband/account APIs are missing, logic safely degrades to character-only completion. 

Why:
Rank remains useful, but newer expansion quest flows are increasingly campaign and account-progress driven. 
This improves post-Shadowlands guide behaviour with a more meaningful completion signal for multi-character players.

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -4565,6 +4565,7 @@ end
 function WoWPro:IsQuestFlaggedCompleted(qid,force)
     if qid == "*" then return nil; end
     local QID = tonumber(qid)
+    local includeWarband = WoWProDB and WoWProDB.profile and WoWProDB.profile.useWarbandCompletion
     if not QID then
         -- is it a QID list?
         local quids = select("#", ("^&"):split(qid))
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -4590,7 +4590,7 @@ function WoWPro:IsQuestFlaggedCompleted(qid,force)
     if not WoWProCharDB.completedQIDs then
         WoWProCharDB.completedQIDs = {}
     end
-    if (WoWProDB and WoWProDB.profile and WoWProDB.profile.useWarbandCompletion) and (not WoWProCharDB.completedQIDsWarband) then
+    if not WoWProCharDB.completedQIDsWarband then
         WoWProCharDB.completedQIDsWarband = {}
     end
 
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -4625,14 +4625,14 @@ function WoWPro:IsQuestFlaggedCompleted(qid,force)
     end
     if QID > 0 then
         if is_int(QID) then
-            return questCompleted(QID)
+            return questCompleted(QID, force)
         else
             QID = floor(QID)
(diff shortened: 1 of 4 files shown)


WoWPro/WoWPro_CompatUtil.lua
```diff
@@ -194,6 +194,20 @@ function WoWPro.QuestLog_IsQuestFlaggedCompleted(questID)
     return _G.C_QuestLog.IsQuestFlaggedCompleted(questID)
 end
 
+function WoWPro.QuestLog_IsQuestFlaggedCompletedOnAccount(questID)
+    if not _G.C_QuestLog then
```
Ludovicus-Maior, line 198: Nice shortcut!

WoWPro/WoWPro_Config.lua
```diff
@@ -132,6 +132,19 @@ local function createDisplayConfig()
...
+                        warbandcompleted = {
+                order = 10.2,
+                type = "toggle",
+                name = L["Use Warband Quest Completion"],
+                desc = L["Treat quests completed by another character in your Warband as completed for step filtering."],
+                get = function(info) return WoWProDB.profile.useWarbandCompletion end,
+                set = function(info,val)
+                    WoWProDB.profile.useWarbandCompletion = val
+                    WoWProCharDB.completedQIDs = {}
+                    WoWProCharDB.completedQIDsWarband = {}
```
Ludovicus-Maior: You should check for other places where completedQIDs is cleared to add the warband equivalent.

Ludovicus-Maior (commented):
A few questions.

WoWPro/WoWPro_Config.lua
Spoony36: Fix is in i think or is there more places you are thinking?

Spoony36:
This can be merged if everyone is happy, Feedback from Cag says its doing what it is supposed to be doing but my play time  is limited these days

Ludovicus-Maior (approved):

WoWPro/WoWPro_Broker.lua
```diff
@@ -4590,31 +4590,49 @@ function WoWPro:IsQuestFlaggedCompleted(qid,force)
     if not WoWProCharDB.completedQIDs then
         WoWProCharDB.completedQIDs = {}
     end
+    if not WoWProCharDB.completedQIDsWarband then
+        WoWProCharDB.completedQIDsWarband = {}
+    end
+
+    local function questCompleted(questID)
```
Furydins: I find functions embedded in other functions make it hard to read and follow the flow.

I'd suggest extracting it out as a local (to the file) method -or if it really has to be embedded moving it to the start of the fdunction.

Spoony36: should be resolved now

PR #3335, opened 2026-05-02
H step completion and verification check

Hendo72:
_* These changes are based on EventConsolidation_

### What changed
- **WoWPro_Broker.lua**
    - Removed `h` from the `C/T` quest-log gating condition.
    - Kept travel/hearth auto-complete logic separate so h is handled by hearth-specific behavior rather than quest-log checks.
    - Added debug logging and `WoWPro.why[...]` text when a travel step is auto-completed/skipped because the player is already in the matching zone.

- **WoWPro_AutoComplete.lua**
    - Added debug log output when a hearthstone bind is saved.
    - Added debug log output when an h step is auto-completed because the hearth location matches the step.
    - Changed hearth persistence from guide-scoped storage (`WoWProCharDB.Guide.hearth`) to global storage (`WoWProCharDB.hearth`).

### Purpose
- Keep `h` steps from being treated like quest complete/turn-in steps.
- Ensure hearth bind state is global and not guide-specific.
- Improve traceability with explicit debug messages for hearth bind and auto-completion events.

_Written with assistance from AI_

### DIFF
--- a/WoWPro/WoWPro_AutoComplete.lua
+++ b/WoWPro/WoWPro_AutoComplete.lua
@@ -398,12 +398,23 @@ function WoWPro:AutoCompleteSetHearth(event, loc, noUpdate)
 
     WoWProCharDB.hearth = loc
     WoWPro:dbp("AutoCompleteSetHearth: hearth bound to [%s] globally via event [%s]", loc, tostring(event))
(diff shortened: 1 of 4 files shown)


Hendo72:
I didn't mean to push those changes to this PR

PR #3340, opened 2026-05-07
Updated to remove Non TBC Quests portals etc

jonv4n-arch:
Removed quests not in TBC, like the Warcheif's command, and the incorrect information about flying mounts, 

The new step one might be better suited with a flight to Stonard, then the note, and some way points to the portal, but I'm not that familiar with the syntax of the guides to be doing that sort of stuff


Hendo72:
I'd make the N step an R step seeing as it's a travel step. Everything looks ok. It'll be a bit before I'm there.

Ludovicus-Maior:
@Hendo72 Still heading in this direction?

Ludovicus-Maior:
@Hendo72 This is a third party pull request for some TBC stuff.  I think it just needs your review and approval.

PR #3342, opened 2026-05-08
Event Code Consolidation plus BONUS Punt/Replay (Reopen)

Ludovicus-Maior:


Ludovicus-Maior:
@Spoony36 Posted the Raptor review:
```
Overall Assessment
Solid architectural improvement. The Punt/Replay system elegantly solves the race condition where events arrive before the guide is fully loaded. Moving to HEARTHSTONE_BOUND over CHAT_MSG_SYSTEM string parsing is the right approach. Still Draft status with a few real issues.

Bugs
1. select(2, ...) in HEARTHSTONE_BOUND handler — off-by-one
In WoWPro_Events.lua (~line 833):

HEARTHSTONE_BOUND fires with a single payload arg (the location string). In the handler, event is already extracted by the dispatch, so ... = (location_string). select(2, ...) on a 1-element vararg returns nothing, so loc is always nil and falls back to GetBindLocation(). It works due to the fallback, but select(2, ...) should be just ... (or select(1, ...)).

2. HEARTHSTONE_BOUND not guarded by WoWPro.RETAIL
HEARTHSTONE_BOUND is a retail-only event. The handler is registered unconditionally, but registering a non-existent event on Classic/TBC/Vanilla can throw a Lua error. Needs if WoWPro.RETAIL then ... end guard (or use RegisterModernEventHandler).

3. WoWPro.GuideUpdated only set on first update
Old code set WoWPro.GuideUpdated = true unconditionally at the end of every UpdateGuideReal() call. New code only sets it inside if not WoWPro.GuideUpdated. Any consumer that checks GuideUpdated to detect a freshly-completed update will see stale true on subsequent updates. Worth auditing all reads of GuideUpdated — though LoadGuide() correctly resets it to false.

Design Notes
ZONE_CHANGED_NEW_AREA direct invocation in UpdateGuideReal:

The comment calls it "Naughty Broker" — fair. Passing a fake event name into a handler that might dispatch on it is fragile. Is there a reason it can't just call the underlying logic function directly?

Production debug print calls in EventHandler (the QuestDialogActive blocks at lines ~171–179): these call WoWPro:print(...) (which appears to be a non-dbp print, visible to all users) every time certain dialogs are open. Intentional or leftover debug?

RegisterAllEvents behavior change: Old version called WoWPro.EventFrame:RegisterAllEvents() (literally subscribes to every WoW event — almost certainly a bug/dead code). New version calls WoWPro:RegisterEvents(nil) which is correct. Good fix but worth noting in the PR since it's a silent behavioral change.

What's Good
Punt/Replay using C_Timer.After + QuestEngineDelay/4 is clean; no OnUpdate polling
Duplicate event registration guard in RegisterEventHandler is solid defensive code
WoWProDB.char.hearth persistence across guides is correct (hearth is character-wide, not guide-scoped)
Clearing EventQueue in both LoadGuide() and OnDisable() covers the edge cases
std = "lua51" in .luacheckrc is the right fix
WoWPro_Frames.lua indentation fix is correct (no behavior change)
InitializeHearthBind() called on both PLAYER_LOGIN and PLAYER_ENTERING_WORLD handles login + zone transitions
Summary of Issues
Severity	Issue
Bug	select(2, ...) → should be ... in HEARTHSTONE_BOUND handler
Bug	HEARTHSTONE_BOUND needs retail-only guard
Watch	GuideUpdated semantics changed — audit consumers
Design	ZONE_CHANGED_NEW_AREA fake-event invocation is fragile
Cleanup	QuestDialogActive print blocks in EventHandler look like debug leftovers
```
Issue #1 is valid and was addressed in https://github.com/Ludovicus-Maior/WoW-Pro-Guides/pull/3342/commits/245709ba3a96cca70e7ac6e37b933506ed5192fb
Issue #2 is not valid.
Issue #3 is will be addressed later.  The case when the user switches guides, but does not reload the addon is where this comes into play!
The debug print calls in EventHandler are a leftover and will be taken care of later.

PR #3347, opened 2026-05-10
Fix TitleBar Collapse and Expanded Isssue

Spoony36:
When the guide frame was collapsed via double-click, any game event (e.g. player movement) that triggered UpdateGuideReal would fall through without returning, causing autoresize/RowSizeSet to expand MainFrame while GuideFrame remained hidden.

- Add `return` after the punt in UpdateGuideReal when GuideFrame is not visible, so the update exits cleanly instead of running layout
- Add `WoWPro.UserCollapsed` flag set on double-click collapse/expand to suppress re-queuing of guide updates while intentionally collapsed

Please confirm this is the correct fix and Test!


Ludovicus-Maior (approved):

Furydins (approved):
LGTM :-)

Ludovicus-Maior:
Weeeee!

PR #3352, opened 2026-05-16
Use IsInGroup for group-sync checks and fix group note gating

Spoony36:
- Replaced legacy group member count checks with IsInGroup for clearer, category-aware group detection.
- Updated group-sync/event gating logic to use home-category checks in WoWPro_Events.lua:413 and WoWPro_Events.lua:443.
- Updated addon group info send guard in WoWPro.lua:925.
- Updated quest share and group-step gating in WoWPro_Broker.lua:1480, WoWPro_Broker.lua:2532, and WoWPro_Broker.lua:2539. 
- Fixed logic bug where the N-step group check was always true due to a >= 0 condition. No functional changes outside group membership detection and related gating.

### DIFF
--- a/WoWPro/WoWPro.lua
+++ b/WoWPro/WoWPro.lua
@@ -922,7 +922,7 @@ end
 
 function WoWPro:SendGroupInfo()
 	--if WoWProCharDB.GroupSync then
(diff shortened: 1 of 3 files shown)


Furydins (approved):
Looks good to me :-)

PR #3354, opened 2026-05-16
Refresh guide window after PLAYER_ENTERING_WORLD

Spoony36:
Potential fix for the issue @Furydins reported here a while back

https://discord.com/channels/195531185403854848/1323715278756118538

Please validate before commiting


Ludovicus-Maior:
Hmm, @Spoony36, with the new event system reworks, it should already be guaranteed that PLAYER_ENTERING_WORLD will result in a guide window update after a delay.   If I am wrong, let me know.

PR #3355, opened 2026-05-16
Fix WoWPro tooltip layering and sizing in guide rows

Spoony36:
- Keep action/row tooltips above guide header text by using dialog strata and frame level.

- Prevent icon tooltip text truncation by auto-sizing width from content with min/max bounds.

- Clamp tooltip to screen to avoid off-screen clipping near edges.

- Preserve existing hover behavior while improving readability and visual priority.

### DIFF
--- a/WoWPro/WoWPro_Widgets.lua
+++ b/WoWPro/WoWPro_Widgets.lua
@@ -58,7 +58,7 @@ function WoWPro:CreateAction(parent, anchor)
     action.frame = frame
     action:SetAllPoints()
 
-    local tooltip = _G.CreateFrame("Frame", nil, frame, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
+    local tooltip = _G.CreateFrame("Frame", nil, _G.UIParent, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
     tooltip:SetBackdrop( {
         bgFile = [[Interface\CHARACTERFRAME\UI-Party-Background]],
         edgeFile = [[Interface\Tooltips\UI-Tooltip-Border]],
@@ -69,7 +69,10 @@ function WoWPro:CreateAction(parent, anchor)
     tooltip:SetHeight(125)
     tooltip:SetWidth(64)
     tooltip:SetAlpha(1)
-    tooltip:SetFrameStrata("TOOLTIP")
+    tooltip:SetFrameStrata("FULLSCREEN_DIALOG")
+    tooltip:SetFrameLevel(1000)
+    tooltip:SetToplevel(true)
+    tooltip:SetClampedToScreen(true)
     tooltip:Hide()
 
     local tooltiptext = tooltip:CreateFontString(nil, nil, "GameFontNormal")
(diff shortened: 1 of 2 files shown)


WoWPro/WoWPro_Widgets.lua
```diff
@@ -69,7 +69,10 @@ function WoWPro:CreateAction(parent, anchor)
     tooltip:SetHeight(125)
     tooltip:SetWidth(64)
     tooltip:SetAlpha(1)
-    tooltip:SetFrameStrata("TOOLTIP")
```
Ludovicus-Maior, line 72: I'm not happy with this change, since if something is being obscured while in the `TOOLTIP` Strata, then the obscuring object is in the wrong strata.

Ludovicus-Maior (changes requested):

WoWPro/WoWPro_Widgets.lua
Spoony36: Latest commit better?

Spoony36:
ok think it's fixed now for real, please check

image

image

image

WoWPro/WoWPro_Widgets.lua
```diff
@@ -58,7 +58,7 @@ function WoWPro:CreateAction(parent, anchor)
     action.frame = frame
     action:SetAllPoints()
 
-    local tooltip = _G.CreateFrame("Frame", nil, frame, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
+    local tooltip = _G.CreateFrame("Frame", nil, _G.UIParent, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
```
Ludovicus-Maior, line 61: Why change the parent frame to the root instead of the nearest frame?

WoWPro/WoWPro_Widgets.lua
```diff
@@ -69,24 +69,31 @@ function WoWPro:CreateAction(parent, anchor)
     tooltip:SetHeight(125)
     tooltip:SetWidth(64)
     tooltip:SetAlpha(1)
-    tooltip:SetFrameStrata("TOOLTIP")
+    tooltip:SetFrameStrata("DIALOG")
```
Ludovicus-Maior, line 72: I don't think this change is needed now.

WoWPro/WoWPro_Widgets.lua
```diff
@@ -574,8 +581,9 @@ function WoWPro:CreateGuideRow(parent, rowHeight)
     local row = _G.CreateFrame("Frame", nil, parent, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
     local tooltip = WoWPro.GuideRowTooltip
     local tooltiptext = tooltip.tooltiptext
-    tooltip:SetParent(parent)
-    tooltip:SetFrameLevel(row:GetFrameLevel() + 1)
+    tooltip:SetParent(_G.UIParent)
```
Ludovicus-Maior, line 584: I think leaving tooltips's parents as they were and letting the Blizzard defaults as they are will work better in the long run.

Ludovicus-Maior (changes requested):

PR #3357, opened 2026-05-16
Improve event handling for F steps

Hendo72:
### Concern
- If the user takes a flight and chooses a destination different from the intended one, or gets off early, the step should not autocomplete until they arrive at the correct destination.
- In its current flow, completion can happen as soon as they board, and the final destination is never confirmed.
- If Auto-completion is triggered too early, `NextStep()` may skip over a step and the guide window starts doing funky things until it's finally sorted out.

### Proposal
- Keep the current `TAXIMAP_OPENED` auto-select behavior.
- Change the taxi trigger flow so completion is based on the correct event transition:
    - detect boarding on PLAYER_CONTROL_LOST + UnitOnTaxi("player")
    - mark the taxi step pending
    - complete it on PLAYER_CONTROL_GAINED once UnitOnTaxi("player") is false

### Changes
1. Add a helper:
    - `WoWPro_IsTaxiStep(qidx)` -> returns `action == "F"`
2. On `PLAYER_CONTROL_LOST`:
    - if current step is an `F` and `_G.UnitOnTaxi("player") is `TRUE`, then set `WoWPro.TaxiPendingStep = qidx`
3. On `PLAYER_CONTROL_GAINED`:
    - if `WoWPro.TaxiPendingStep` exists, current `qidx` matches it, and `_G.UnitOnTaxi("player")` is `FALSE`, complete that pending step and clear `WoWPro.TaxiPendingStep`
    
    **Note:** 
This is a trigger correction, not a change to the higher-level auto-completion policy. The completion still happens, but now it happens at the right trigger point.
This also prevents flights that take the scenic route and fly over a destination early from triggering before it's ready.

**I almost missed this:**
For boat/ship travel, the same bail-out risk exists, but boat don't have a taxi interface and are not covered by this taxi-event fix. No more different than running to a different location.

### DIFF
--- a/WoWPro/WoWPro_Events.lua
+++ b/WoWPro/WoWPro_Events.lua
@@ -801,7 +801,7 @@ end)
 WoWPro.RegisterEventHandler("TAXIMAP_OPENED", function(event, ...)
     WoWPro:RecordTaxiLocations(...)
     local qidx = WoWPro.rows[WoWPro:GetActiveStickyCount()+1].index
-    if (WoWPro.action[qidx] == "F" or WoWPro.action[qidx] == "b") then
+    if WoWPro.action[qidx] == "F" then
         if WoWProCharDB.AutoSelect == true then
             WoWPro.TakeTaxi(WoWPro.step[qidx])
         else
@@ -819,7 +819,7 @@ end)
 
 function WoWPro.PLAYER_CONTROL_LOST_PUNTED(event, ...)
     local qidx = WoWPro.rows[WoWPro:GetActiveStickyCount()+1].index
-    if (WoWPro.action[qidx] == "F" or WoWPro.action[qidx] == "b") then
+    if WoWPro.action[qidx] == "F" then
         if _G.UnitOnTaxi("player") then
             WoWPro.TaxiPendingStep = qidx
             WoWPro:dbp("PLAYER_CONTROL_LOST_PUNTED: UnitOnTaxi! pending taxi completion for step %d", qidx)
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_Events.lua
```diff
@@ -801,7 +801,7 @@ end)
 WoWPro.RegisterEventHandler("TAXIMAP_OPENED", function(event, ...)
     WoWPro:RecordTaxiLocations(...)
     local qidx = WoWPro.rows[WoWPro:GetActiveStickyCount()+1].index
-    if (WoWPro.action[qidx] == "F" or WoWPro.action[qidx] == "b") then
```
Ludovicus-Maior, line 804: This eliminates the handling for boats (which the client treats as a type of taxi) which is not good.
Changing every WoWPro.action[qidx] == "F" back to a (WoWPro.action[qidx] == "F" or WoWPro.action[qidx] == "b") would make me happier.

WoWPro/WoWPro_Events.lua
```diff
@@ -819,16 +819,33 @@ end)
...
-            WoWPro.CompleteStep(qidx,"Took a taxi")
+            WoWPro.TaxiPendingStep = qidx
+            WoWPro:dbp("PLAYER_CONTROL_LOST_PUNTED: UnitOnTaxi! pending taxi completion for step %d", qidx)
         else
             WoWPro:dbp("PLAYER_CONTROL_LOST_PUNTED: not on taxi!")
         end
     end
 end
 
+WoWPro.RegisterEventHandler("PLAYER_CONTROL_GAINED", function(event, ...)
```
Ludovicus-Maior, line 835: This is a good change.  Do we need the `PLAYER_CONTROL_LOST_PUNTED` handler and whoever punts it anymore?

Ludovicus-Maior (changes requested):

WoWPro/WoWPro_Events.lua
Hendo72: I changed it back and took it one step further and included R steps where the mount is cinematic.

WoWPro/WoWPro_Events.lua
Hendo72: It adds a delay based on the event.

WoWPro/WoWPro_Events.lua
Hendo72: I fixed it and took it one step further by adding R steps for the edge case where the travel is by cinematic
- the cat in Darkshore, for example

Ludovicus-Maior: Ohhh, good catch!

WoWPro/WoWPro_Events.lua
Ludovicus-Maior: Sounds good enuf to me!

Ludovicus-Maior (approved):

PR #3358, opened 2026-05-20
Auto complete bug fixes

Hendo72:
### Fix
- Removed a duplicate allPositiveComplete branch from WoWPro_Broker.lua
    - The same loot completion logic was being executed twice
    - That could cause duplicate state changes, duplicate CompleteStep calls, and confusing behavior
- Fixed wrong comparator logic for quest and scenario objective operators
    - QuestLess now means “objective count is less than or equal to target”
    - QuestGreater now means “objective count is greater than or equal to target”
    - ScenarioLess now means “scenario quantity is less than or equal to target”
    - ScenarioGreater now means “scenario quantity is greater than or equal to target”
- Added a comment documenting the corrected operator meaning

### The Issue
- The duplicate branch made loot completion logic run twice for the same condition
- The comparator functions were backwards compared to Q<, Q>, S<, and S>
- That mismatch made guide conditions harder to understand and could produce incorrect step behavior

### DIFF
--- a/WoWPro/WoWPro_AutoComplete.lua
+++ b/WoWPro/WoWPro_AutoComplete.lua
@@ -224,6 +224,29 @@ function WoWPro.GetLootTrackingInfo(lootItems)
     return tracks
 end
 
+function WoWPro.LootItemsCollected(lootItems)
+    if not lootItems or next(lootItems) == nil then
+        return true
+    end
+    local allPositiveComplete = true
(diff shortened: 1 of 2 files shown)


Hendo72:
@Spoony36 Have a look and tell me what you think.
@Ludovicus-Maior @Furydins This isn't in the memory leak refactor I'm working on

PR #3359, opened 2026-05-21
Feature/broker onupdate memory opt

Hendo72:
## TL;DR
1. **Consolidate and stabilize broker refresh timing**
    - use one shared updater instead of many separate row refresh handlers.
    - one timer controls broker row updates, so refreshes happen in one place and at a steady pace.
2. **Make row refresh safe during combat**
    - detect combat lockdown and postpone UI changes that are not allowed while fighting.
    - don’t touch protected broker row state in combat; wait until it’s safe and then update cleanly.
3. **Cut repeated table allocations from `RowUpdate()`**
    - keep four reusable step arrays on WoWPro and clear them with `table.wipe(...)` instead of recreating them.
    - reuse the same temporary buffers every refresh so the addon does not make new garbage tables over and over.

#### Why:
- The old update model was scattered and inefficient. Each row could end up managing its own refresh behavior, and updates could happen unevenly.
#### How:
- The branch centralizes refresh work into one shared broker updater frame:
    - `BrokerUpdateFrame:SetScript("OnUpdate", ...)`
    - one fixed-timer loop drives row refreshes
    - rows register/unregister with that shared updater
#### Result:
- more consistent broker updates, easier maintenance, and a single place to control refresh timing.

#### Why:
- WoW UI is strict during combat. Changing protected rows or button bindings while in combat can cause errors or break the addon.
#### How:
- The branch adds safety checks and deferred handling:
    - `WoWPro.MaybeCombatLockdown()` blocks unsafe refresh activity during combat
    - row updates that touch secure state are delayed until combat ends
    - a “flush after combat” helper applies pending row changes safely once combat is over
    - override bindings are cleared only when not in combat
#### Result:
- broker refresh behaves correctly in combat and does not trigger protected-frame failures.

#### Why:
- `RowUpdate()` runs frequently, and allocating temporary tables every refresh creates a lot of garbage. That leads to Lua GC pauses and UI stutter.
#### How:
 - The branch reuses internal scratch buffers instead of rebuilding them each time:
    - replace temporary tables with persistent ones on WoWPro
    - `WoWPro.RowUpdateAllSteps`
    - `WoWPro.RowUpdateStickySteps`
    - `WoWPro.RowUpdateRegularSteps`
    - `WoWPro.RowUpdateStepList`
    - each update now uses table.wipe(...) instead of {}
#### Result:
 - the same row selection and sorting logic runs, but with much lower allocation overhead and smoother refresh performance.

### DIFF
--- a/WoWPro_Leveling/TBC/Horde/INTRO_BCC_Tauren.lua
+++ b/WoWPro_Leveling/TBC/Horde/INTRO_BCC_Tauren.lua
@@ -7,292 +7,292 @@ WoWPro:GuideLevels(guide, 1, 12, 2)
 WoWPro:GuideNextGuide(guide, 'CLASSIC_BC_Silverpine_Forest')
 WoWPro:GuideSteps(guide, function() return [[
 
-A The Hunt Begins|QID|747|M|44.87,77.08|N|From Grull Hawkwind.|
-A A Humble Task|QID|752|M|44.19,76.06|N|From Chief Hawkwind.|
-C The Hunt Begins|QID|747|M|47.87,77.76|QO|1|N|Kill Plainstriders around the camp.|
+A The Hunt Begins|QID|747|M|44.87,77.08|Z|1412; Mulgore|N|From Grull Hawkwind.|
+A A Humble Task|QID|752|M|44.19,76.06|Z|1412; Mulgore|N|From Chief Hawkwind.|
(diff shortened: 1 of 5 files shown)


WoWPro_Leveling/TBC/Horde/INTRO_BCC_Tauren.lua
Hendo72, line 1: This wasn't meant to be included. I was working on the guide while testing.

Hendo72:
I thought this was already taken care of.

This is the lag fix I was exploring.

I assumed that the conflicts meant it was.

PR #3365, opened 2026-05-27
TZ tag changes

Cagomei:
Requested by Hendo to add the Zone name to them as well.


Ludovicus-Maior:
Guide changes to use a pre-existing feature.  
Misspellings are amusing and not dangerous.

PR #3367, opened 2026-05-28
Fix US step visibility filtering for sticky step flow

Hendo72:
**Bug**
- `US` steps were being treated like they were skipped when their paired `S` step was still incomplete.
- In WoWPro_Broker.lua, the regular step list construction removed `US` steps unless the matching `S` step was already completed.
- That meant a `US` could disappear from the guide even though it should still exist, waiting for earlier rows to clear.

**Why it was wrong**
- The `US` visibility decision is a display filter, not a skip condition.
- `US` should only be hidden when it is not the first non-sticky row after active stickies.
- The matching/completion logic for `S` was correct, but the visibility gate blocked the `US` too early.

**Fix**
- Change the step list merge logic so:
    - all sticky steps are added first
    - regular steps are added in order
    - `US` steps are only included if they are the first regular step after the stickies
    - This preserves intended UI behavior:
        - if a `US` is the first eligible regular row, it is shown
        - later `US` rows remain hidden until earlier regular rows are done

**Result**
- US now becomes active at the right time
- when it reaches `ActiveStickyCount + 1`, it can complete its paired S
- display/filtering now matches the intended `S`/`US` semantics instead of confusing it with skip logic

### DIFF
--- a/WoWPro/WoWPro.lua
+++ b/WoWPro/WoWPro.lua
@@ -1220,13 +1220,6 @@ WoWPro.MIDNIGHT = ((WoWPro.TocVersion >= 120000) and (WoWPro.TocVersion < 130000
 WoWPro.RETAIL_RELEASE = 12
 WoWPro.RETAIL = (WoWPro.TocVersion >= WoWPro.RETAIL_RELEASE * 10000)
 
--- TEMP DEBUG: Verify retail/runtime classification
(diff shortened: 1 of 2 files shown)


Hendo72:
Do not merge this until i can identify where the lag was introduced.

Hendo72:
This PR is screwed. I'm scrapping and restarting it.

PR #3374, opened 2026-06-05
Fix stale activestep refresh after quest completion

Hendo72:
This change fixes a case where a quest step could complete successfully, but the guide would continue showing the old step until the user manually reset or reloaded it.

The issue was not that the quest completion failed. The completion was recorded, but the guide did not always refresh immediately afterward, which left the UI looking stuck even though progress had already advanced behind the scenes.

To address this, the completion flow now triggers an immediate guide refresh when the completed step is also the current active step, instead of waiting for the normal deferred update path. That immediate refresh only runs when it is safe to do so, so the existing behavior remains unchanged in combat or other restricted states.

A small always-on log breadcrumb was also added for this bypass path. This gives us a clear signal in future reports without requiring debug logging and without adding extra noise to normal logs.


Hendo72:
The second commit is an expansion on the log message to help with future issues surrounding this problem. I separated this to make dismissing it easier. I think it might be helpful.

PR #3375, opened 2026-06-06
No more choppy chopy

Hendo72:
** Not sure if this different than the fix that Spoony presented earlier.

**Bug**
- The tooltip was being drawn inside a box that cuts off anything that sticks out.
- So when the tooltip appeared near the top, its top got chopped.

**What we changed:**
- We moved the tooltip to the main screen layer (not inside that cutting box).
- We still make it appear next to the step icon.
- If the guide is near the top of the screen, we show the tooltip below the icon instead of above it.

**So now:**
- It can extend past the guide window without being clipped.
- It won’t get cut off at the top.

_Researched and prepared using GPT-5-3-Codex_

### DIFF
--- a/WoWPro/WoWPro_Widgets.lua
+++ b/WoWPro/WoWPro_Widgets.lua
@@ -549,7 +549,7 @@ function WoWPro:CreateScrollbar(parent, offset, step, where)
 end
 
 do
-    local tooltip = _G.CreateFrame("Frame", nil, _G.UIParent, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
+    local tooltip = _G.CreateFrame("Frame", nil, WoWPro.MainFrame, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
     tooltip:SetBackdrop( {
         bgFile = [[Interface\Tooltips\UI-Tooltip-Background]],
         edgeFile = [[Interface\Tooltips\UI-Tooltip-Border]],
--- a/WoWPro/WoWPro_Widgets.lua
+++ b/WoWPro/WoWPro_Widgets.lua
@@ -48,7 +48,7 @@ function WoWPro:CreateCheck(parent)
     return check
 end
 
-function WoWPro:CreateAction(parent, anchor)
+function WoWPro:CreateIcon(parent, anchor)
     local frame = _G.CreateFrame("Frame", nil, parent)
     frame:SetPoint("LEFT", anchor, "RIGHT", 3, 0)
     frame:SetWidth(16)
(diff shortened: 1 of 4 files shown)


WoWPro/WoWPro_Widgets.lua
```diff
@@ -53,12 +53,13 @@ function WoWPro:CreateAction(parent, anchor)
...
     frame:SetWidth(16)
     frame:SetHeight(16)
+    frame:EnableMouse(true)
 
     local action = frame:CreateTexture()
     action.frame = frame
     action:SetAllPoints()
 
-    local tooltip = _G.CreateFrame("Frame", nil, frame, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
+    local tooltip = _G.CreateFrame("Frame", nil, _G.UIParent, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)
```
Spoony36: This is incorrect and why mine hasn't been merged yet.

Furydins:
It is the same bug!

Ludovicus-Maior:
@Spoony36 Since your name was brought up in the pull request,  I would like your review.

Spoony36:
> @Spoony36 Since your name was brought up in the pull request, I would like your review.

Not sure why @Hendo72  has changed so much code but i 'think' this will suffice at least to have the tooltip show.

 Hendo can you test this code change or explain why you are changing so much code, might be a valid reason for your code changes but for the tooltip i think this will be ok?

image

fix: tooltip no longer obscured by sibling frames

The action icon tooltip was parented to the 16x16 icon frame child,
constraining its draw order to that frame's level. Re-parent it to the
row frame (the nearest ancestor passed as `parent`) so its frame level
sits above guide row siblings. SetFrameStrata("TOOLTIP") then reliably
renders it above all other UI.

Also widen tooltip from 64px to 160px (text from 50px to 140px) so
labels such as "Campaign Quest" are not clipped, and add screen-edge
awareness to flip the anchor left when the tooltip would overflow the
right side of the screen.

Spoony36:
or even instead of parent

local tooltip = _G.CreateFrame("Frame", nil, WoWPro.MainFrame, _G.BackdropTemplateMixin and "BackdropTemplate" or nil)

The tooltip is now parented to WoWPro.MainFrame. That's the right level — it escapes GuideFrame's SetClipsChildren(true) clipping region, but stays within the addon's own frame hierarchy rather than going all the way to UIParent

WoWPro/WoWPro_Widgets.lua
Ludovicus-Maior: In general, all frames an application creates should be anchored to a frame hierarchy identifiable to the application.  `tooltips` should be anchored to the object that should have the tooltip displayed when you move the mouse into it.  The only exception to these rules, are global dialog boxes, which are anchored to UIParent by the Blizzard API itself.  This convention makes it possible for `/fstack` to correcly identify the owners of all the UI elements on the screen for debugging. We could legitimately be publicly spanked by others by violating this convention.

Ludovicus-Maior:
Hmmpf!  Newfangled API calls...

PR #3378, opened 2026-06-10
Previous Recursive fix wasn't enough

Hendo72:
I fixed a bug in WoWPro_Broker.lua where completing a step could call `UpdateGuideReal()` while the guide was already updating, which caused a recursive loop and overflow.

The previous fix added a guard, but the problem was that `CompleteStep()` could still choose the direct `UpdateGuideReal()` path too often. That fast path can still be dangerous if the guide is already in the middle of updating, so the fix is to only do it when it is clearly safe and otherwise fall back to the normal queued update.

The code now only does an immediate refresh when it is safe to do so, and it keeps the existing `UpdateGuideRealInProgress` guard to stop nested updates. This should prevent the guide from recursing and crashing during step completion.

***
** I'm also tacking on a fix for a bug that was discovered while researching the original issue.

**Root cause:** 
- Decimal QIDs like `1234.1` or `1234.2` could be checked using the wrong internal key.
- That means the addon might think a quest step is completed when it is not, or not completed when it is.
- Guides using decimal QIDs could fail to advance correctly or skip steps incorrectly.
- **In short:** fractional QIDs would report the wrong completion status.

**10 years in hiding**
- If a guide uses plain integer QIDs or the player never reaches that exact step, the bug stays hidden.
- Also, the wrong result can be subtle: the guide may still work most of the time, or it may only skip one step.
- it was a rare, edge-case bug tied to a special QID form, so it could easily go unnoticed.

**Fix**
- In `WoWPro:IsQuestFlaggedCompleted(qid, force)` the code now detects positive decimal QIDs like `1234.1`.
- It uses` -floor(QID)` as the cache key instead of trying to cache the raw decimal.
- For a decimal QID, it checks completion using the base quest ID `(floor(QID))` and stores the result under that negative base key.

_So the fix is_: when QID is positive and not an integer, normalize it by `-floor(QID)` before caching and checking completion.

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -18,6 +18,8 @@ WoWPro.mygroupsteps = {}
 WoWPro.myGroupTrack = {}
 WoWPro.playerGroup = {}
 
+WoWPro.UpdateGuideRealInProgress = false
+
 -- Debug toggles
 WoWPro.DEBUG_STICKY_PAIRING = false -- Set to true to enable sticky pairing debug output
 WoWPro.DEBUG_REPEATABLE = false -- Set to true to enable debug output for repeatable A step resets and quest log changes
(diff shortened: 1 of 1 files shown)


Ludovicus-Maior:
I'm not sure this is correct.
Decimal QUIDs are used to trigger fake quest accepts when a given pet NPC, for example in the garrison, is moused over.  I'm not 100% sure, but I think all decimal quests are informational and I know they are only supposed to be used for pet battle quest selection.   Once the NPC to QID quest mapping table is set up, the step should be marked as completed at guide load time.

UpdateGuide should not be called during guide load.

Ludovicus-Maior:
This PR is superseded by https://github.com/Ludovicus-Maior/WoW-Pro-Guides/pull/3379 .

PR #3379, opened 2026-06-10
Previous Recursive fix wasn't enough (And too much!)

Ludovicus-Maior:
In PR https://github.com/Ludovicus-Maior/WoW-Pro-Guides/pull/3378 @Hendo72 fixed a recursion error, but the silly AI thingy got spooked by something that was not broken.  This is the same PR, sans the extraneous fix.

Hendo:
I fixed a bug in WoWPro_Broker.lua where completing a step could call `UpdateGuideReal()` while the guide was already updating, which caused a recursive loop and overflow.

The previous fix added a guard, but the problem was that `CompleteStep()` could still choose the direct `UpdateGuideReal()` path too often. That fast path can still be dangerous if the guide is already in the middle of updating, so the fix is to only do it when it is clearly safe and otherwise fall back to the normal queued update.

The code now only does an immediate refresh when it is safe to do so, and it keeps the existing `UpdateGuideRealInProgress` guard to stop nested updates. This should prevent the guide from recursing and crashing during step completion.

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -18,6 +18,8 @@ WoWPro.mygroupsteps = {}
 WoWPro.myGroupTrack = {}
 WoWPro.playerGroup = {}
 
+WoWPro.UpdateGuideRealInProgress = false
+
 -- Debug toggles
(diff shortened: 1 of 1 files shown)


Hendo72 (approved):
2 thumbs up

PR #3380, opened 2026-06-11
Fixed raid target marking for Retail & Dragonblight optimization

Saaappi:
Raid target marking no longer supports a pre-check in Retail, so implemented just the /tm slash command. Optimized the horde Dragonblight guide a bit.

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -18,8 +18,6 @@ WoWPro.mygroupsteps = {}
 WoWPro.myGroupTrack = {}
 WoWPro.playerGroup = {}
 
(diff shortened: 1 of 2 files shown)


Furydins (approved):

Ludovicus-Maior:
@Saaappi I am going to smash and merge this pull request with https://github.com/Ludovicus-Maior/WoW-Pro-Guides/pull/3381 so that the guide changes and the code changes are clearly separated.  There are a few errors in  the guides that need fixing (repeated tags in steps are not supported and should have been flagged as an error).  There was quite a bit of discussion about the right way to fix them, and it took awhile to reach consensus.  

Thank you for the PR's, nonetheless.

PR #3389, opened 2026-06-17
Make Ludo Panic and Update WoWPro_Broker.lua

Spoony36:
As always confirm validity and test. Feel free to reject if you had a different approach in mind

Marked as draft so others can contribute or edit

- Introduced WoWPro.StepState state enum in WoWPro_Broker.lua 
- Added GetStepState(stepIndex) and SetStepState(stepIndex, state, reason) helpers 
- Replaced direct completion/skipped table mutations in core broker path with the new state helpers 
- Updated CompleteStep() to use SetStepState(..., DONE, reason) 
- Refactored NextStep() to query step status through GetStepState() and to mark skipped steps through state helper 
- Laid groundwork for a broker-driven state machine for step activation, skipping, and completion

- Refactor internal event pump to Blizzard scheduler and preserve direct dispatch semantics

- Reworked WoWPro:PostInternalMessage to queue internal events and schedule processing exactly once using C_Timer.After(0, ...) or fallback OnUpdate
- Added ScheduleInternalMessagePump() helper and stable InternalMessagePump.scheduled state
- Kept WoWPro:ProcessInternalMessages() clearing the queue before dispatch and resetting the scheduled flag
- Retained direct listener invocation in WoWPro:DispatchEvent(eventName, ...) so WoW/Blizzard error handling remains intact
- Clarified that payloads may come from original frame events or internally generated messages, so no extra error-wrapping is required

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -101,16 +101,23 @@ function WoWPro:DispatchEvent(eventName, ...)
 
     for listenerFn, purpose in pairs(snapshot) do
         if listeners[listenerFn] == purpose then
-            local ok, err = pcall(listenerFn, eventName, ...)
-            if not ok then
+            local function dispatchErrorHandler(err)
+                local message = tostring(err)
                 WoWPro:Error("DispatchEvent(%s): listener=%s purpose=%s error=%s",
-                    tostring(eventName), tostring(listenerFn), tostring(purpose), tostring(err))
+                    tostring(eventName), tostring(listenerFn), tostring(purpose), message)
+                local handler = geterrorhandler()
+                if handler then
+                    handler(message)
+                end
+                return message
+            end
+
+            if not xpcall(listenerFn, dispatchErrorHandler, eventName, ...) then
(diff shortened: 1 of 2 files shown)


Spoony36:
not sure why its changing the texture stuff, i didn't touch that and my IDE isn't showing those changes

WoWPro/WoWPro_Broker.lua
```diff
@@ -101,20 +101,7 @@ function WoWPro:DispatchEvent(eventName, ...)
...
-                if handler then
-                    handler(message)
-                end
-                return message
-            end
-
-            if not xpcall(listenerFn, dispatchErrorHandler, eventName, ...) then
-                listeners[listenerFn] = nil
-            end
+            listenerFn(eventName, ...)
```
Ludovicus-Maior, line 148: ¡Mucho mas bueno!

Ludovicus-Maior: Without thinking too hard, I would wager … could contain information either from the original frame event that triggered this or the data structure that indirectly triggered our internal event.

PR #3390, opened 2026-06-17
Emma Request -Make recorder auto tag K steps

Spoony36:
- detect ACTIVE + QO objectives that look like a named mob kill convert those into a K step

- extract the mob name into target

Please check this wont affect anything and will do what i think it will


Ludovicus-Maior:
This is not a perfect strategy. 
However, the recorder needs to guess better and this does that.
The guide writer always needs to review the output and do their best.

PR #3407, opened 2026-06-30
That's not how it works

Hendo72:
The `nil` test was done wrong.

TBH, there's no need to test for nil anyway.

if the event isn't "`HEARTHSTONE_BOUND`", then it doesn't matter anyway; return early.

### DIFF
--- a/WoWPro/WoWPro_AutoComplete.lua
+++ b/WoWPro/WoWPro_AutoComplete.lua
@@ -413,7 +413,7 @@ end
 
 -- Auto-Complete: Set hearth --
 function WoWPro:AutoCompleteSetHearth(event, loc, noUpdate)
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_AutoComplete.lua
```diff
@@ -413,7 +413,7 @@ end
 
 -- Auto-Complete: Set hearth --
 function WoWPro:AutoCompleteSetHearth(event, loc, noUpdate)
-    if event and event ~= "HEARTHSTONE_BOUND" then
+    if event ~= "HEARTHSTONE_BOUND" then
```
Ludovicus-Maior, line 416: Redundant, but harmless.

PR #3414, opened 2026-07-07
Be Quiet: completion sound refactor

Hendo72:
This PR tightens completion-sound behavior in WoWPro so the check sound only plays when the completed step is actually visible in the guide window. The sound decision now flows through a centralized helper in CompleteStep, which keeps the behavior consistent and makes future changes easier to reason about.

I also removed the manual right-click sound bypass and added diagnostics for the main background completion paths, including quest updates, travel, bucketed refreshes, and sticky/unsticky pairing.

Each commit was a progressive advancement in implementing the code. Cherry-pick each commit if you wish to do testing

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -966,9 +966,6 @@ function WoWPro:CheckFunction(row, button, down)
     elseif button == "RightButton" and row.check:GetChecked() then
         row.check:SetGold()
         WoWPro:dbp("WoWPro:CheckFunction: User marked step %d as complete.", row.index)
-        if WoWProDB.profile.checksound then
-            _G.PlaySoundFile(WoWProDB.profile.checksoundfile)
-        end
         -- if CompleteStep() did a LoadGuide, skip out.
         if WoWPro.CompleteStep(row.index,"Right-Click") then
             return
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -96,7 +96,7 @@ local function ShouldPlayCompletionSound(step, noUpdate, alreadyComplete, soundC
     if alreadyComplete then
         return false
     end
-    if noUpdate then
+    if noUpdate and not (soundCtx and soundCtx.allowSoundWhileNoUpdate) then
         return false
     end
     if soundCtx and soundCtx.suppressSound then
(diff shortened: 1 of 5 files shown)


Spoony36 (commented):
Here's an AI review from me, it might be talking utter rubbish but here it is

I like the direction here. Centralizing completion-sound policy in CompleteStep, removing the direct right-click bypass, and adding origin diagnostics all make this behavior easier to reason about.

I do see one blocking issue with the new visibility gate:

ShouldPlayCompletionSound() now depends on IsStepVisibleInGuide(step), which checks the current WoWPro.rows state. But in the sticky/unsticky path, paired-sticky completion still happens before the rows are rebuilt for the current update pass. In UpdateGuideReal(), ActiveStep is chosen and the paired sticky is completed before RowUpdate() runs. The actual visibility/filtering logic that decides whether that sticky should be shown is inside RowUpdate().

That means sound eligibility is being computed from stale row visibility, not from the current logical guide state. A step can still be considered “visible” based on the previous render and play a sound on the same pass where the new row model is about to hide it. That doesn’t fully match the PR’s stated policy of only sounding when the completed step is actually visible in the guide window.

I’d suggest one of these fixes:

Base sound eligibility on the same logical visibility/filter rules used to build the step list in RowUpdate(), instead of asking the reused row frames whether they are currently visible.
Or defer the sound decision until after the guide rows have been rebuilt for the current pass.

Non-blocking improvements:

Replace ad hoc soundCtx.origin strings with shared constants or a small enum-style table.

Add a small test matrix in the PR description covering right-click completion, hidden guide frame, filtered AVAILABLE/ACTIVE steps, sticky→unsticky paired completion, and background autocomplete while the step is offscreen.

Overall: good refactor, but I would not merge it until the visibility gate is tied to current guide state rather than stale row widget visibility.

WoWPro/WoWPro_Broker.lua
```diff
@@ -4169,15 +4247,32 @@ function WoWPro.NextStepNotSticky(guideIndex)
 end
 
 -- Step Completion Tasks --
-function WoWPro.CompleteStep(step, why, noUpdate)
+function WoWPro.CompleteStep(step, why, noUpdate, soundCtx)
```
Ludovicus-Maior: Can the soundCtx dictonary/array ever hve a key except for `origin`.  If not, it really just ought to be called `origin` and be a string and avoid the array construction.

Ludovicus-Maior (changes requested):
Mostly OK.  I don't like the `{ origin = "foo"}` argument.

WoWPro/WoWPro_Broker.lua
Hendo72: I see what you mean and I've made the appropriate changes.

I did forget to set the one stipulation that I always do and that's the rule of 3. This new commit addresses that as well.

PR #3419, opened 2026-07-14
Update WoWPro_Parser.lua

Spoony36:
Fix unskip cascade QID type comparison in parser

### DIFF
--- a/WoWPro/WoWPro_Parser.lua
+++ b/WoWPro/WoWPro_Parser.lua
@@ -331,7 +331,7 @@ function WoWPro.UnSkipStep(index)
             local numprereqs = select("#", ("^&"):split(WoWPro.prereq[j]))
             for k=1,numprereqs do
                 local kprereq = select(numprereqs-k+1, ("^&"):split(WoWPro.prereq[j]))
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_Parser.lua
```diff
@@ -331,7 +331,7 @@ function WoWPro.UnSkipStep(index)
             local numprereqs = select("#", ("^&"):split(WoWPro.prereq[j]))
             for k=1,numprereqs do
                 local kprereq = select(numprereqs-k+1, ("^&"):split(WoWPro.prereq[j]))
-                if tonumber(kprereq) and tonumber(kprereq) == WoWPro.QID[currentstep] then
+                if tonumber(kprereq) and tonumber(kprereq) == tonumber(WoWPro.QID[currentstep]) then
```
Ludovicus-Maior: Not quite right, methinks. 
You need to test to see if `kprereq` is present in `WoWPro.QID[currentstep])`.  The operators do not matter for the first approximation.  UNskip the step and let the Broker fix it up later.

Ludovicus-Maior (changes requested):

PR #3421, opened 2026-07-14
fix: make quest icon selection deterministic

Spoony36:
This fixes three quest-icon helpers in WoWPro_Parser.lua so they no longer rely on pairs() over hash tables. Instead, they now iterate explicit priority lists with ipairs(), which makes icon selection stable when multiple quest flags are set at once

### DIFF
--- a/WoWPro/WoWPro_Parser.lua
+++ b/WoWPro/WoWPro_Parser.lua
@@ -219,7 +219,7 @@ function WoWPro.GetQuestIconComplete(QID)
     end
     return WoWPro.CompleteIcons["default"], "Default"
 end
-end
+
 for flag, icon in pairs(WoWPro.OfferIcons) do
     if string.sub(flag,1,2) == "is" then
         local key = string.format("A %s", string.sub(flag,3))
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_Parser.lua
```diff
@@ -331,7 +360,7 @@ function WoWPro.UnSkipStep(index)
             local numprereqs = select("#", ("^&"):split(WoWPro.prereq[j]))
             for k=1,numprereqs do
                 local kprereq = select(numprereqs-k+1, ("^&"):split(WoWPro.prereq[j]))
-                if tonumber(kprereq) and tonumber(kprereq) == WoWPro.QID[currentstep] then
+                if tonumber(kprereq) and tonumber(kprereq) == tonumber(WoWPro.QID[currentstep]) then
```
Ludovicus-Maior, line 363: Note: This suffers from the fact that WoWPro.QID may not be a number, if it is a QID expression!

Ludovicus-Maior (changes requested):

WoWPro/WoWPro_Parser.lua
Spoony36: Will close and revisit as a few changes have been made since and now a conflict

PR #3422, opened 2026-07-14
reject malformed operator objectives in validation

Spoony36:
reject malformed operator objectives in validation

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -367,11 +367,23 @@ function WoWPro:QuestFailed(QIDs, debug, why)
 end
 
 local OBJECTIVE_PATTERN = "^(%d*)([<=>]*)(%d*)$"
+local function CanonicalizeObjectiveOperator(operator)
+    if operator == "<=" then
+        return "<"
+    elseif operator == ">=" then
+        return ">"
+    elseif operator == "<" or operator == ">" or operator == "=" or operator == "" then
+        return operator
+    end
+    return nil
+end
+
 function WoWPro.ValidObjective(questtext)
     local objective, operator, target = tostring(questtext):match(OBJECTIVE_PATTERN)
+    local canonical = CanonicalizeObjectiveOperator(operator)
     if operator == "" and target == "" then
         return tonumber(objective)
-    elseif operator ~= "" and target ~= "" and operator:match("^[<=>]$") then
+    elseif operator ~= "" and target ~= "" and canonical then
         return tonumber(objective)
     else
         return false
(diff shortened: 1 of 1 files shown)


WoWPro/WoWPro_Broker.lua
```diff
@@ -369,7 +369,7 @@ end
 local OBJECTIVE_PATTERN = "^(%d*)([<=>]*)(%d*)$"
 function WoWPro.ValidObjective(questtext)
     local objective, operator, target = tostring(questtext):match(OBJECTIVE_PATTERN)
-    if operator ~= "" then
+    if operator ~= "" and target ~= "" then
```
Ludovicus-Maior: I would have thought his would have been an `or` expression and not an `and`.

Ludovicus-Maior (changes requested):

WoWPro/WoWPro_Broker.lua
Spoony36: You’re right to question. I dug through the flow and the real issue was broader: even with or, malformed objectives could still pass because the fallback branch treated empty strings as truthy in Lua.

I updated validation so it now only accepts:

plain numeric objective (no operator, no target), or
single valid operator (<, =, >) with a non-empty numeric target.
So malformed forms like 1<, <3, or 1<>2 are now rejected at validation, which keeps it consistent with parser behavior.

WoWPro/WoWPro_Broker.lua
```diff
@@ -369,12 +369,10 @@ end
 local OBJECTIVE_PATTERN = "^(%d*)([<=>]*)(%d*)$"
 function WoWPro.ValidObjective(questtext)
     local objective, operator, target = tostring(questtext):match(OBJECTIVE_PATTERN)
-    if operator ~= "" then
+    if operator == "" and target == "" then
         return tonumber(objective)
-    elseif target == "" and operator == "" then
+    elseif operator ~= "" and target ~= "" and operator:match("^[<=>]$") then
```
Ludovicus-Maior: So what about `<=` ?

Ludovicus-Maior (changes requested):

WoWPro/WoWPro_Broker.lua
Spoony36: Support <= and >= objective operators in objective parsing

Description:

Add objective-operator canonicalization so <= maps to < and >= maps to >.
Apply canonicalized operators in both objective validation and objective parsing to keep behavior consistent.
Preserve existing malformed-operator handling, still falling back safely when parsing invalid objective expressions.
This addresses the PR review gap around <= support without changing existing comparator semantics.

PR #3425, opened 2026-07-16
Fix combat lockdown frame dragging

Hendo72:
Guard all `StartMoving()` calls in WoWPro_Frames.lua with `not InCombatLockdown()` to prevent protected function errors while in combat. This prevents accidental drag attempts from triggering [ADDON_ACTION_BLOCKED] when the user clicks the frame during combat.


Ludovicus-Maior:
Ugggh.  These new combat lockdown restrictions are effing annoying.

PR #3434, opened 2026-08-02
simplify profession data handling and harden tradeskill scanning

Spoony36:
This change consolidates profession/tradeskill handling across the earlier refactor commits and the latest follow-up cleanup.

It replaces duplicated profession skill-line tables with data-driven builders, introduces an explicit no-window profession refresh path, and keeps backward compatibility by preserving the old update entry point as an alias.

It also strengthens scan reliability and event behavior:

Rewires startup and skill-line events to use explicit current-profession refresh.
Adds debounced handling for bursty skill-line updates.
Refreshes profession data on LEARNED_SPELL_IN_SKILL_LINE.
Preserves detailed, window-dependent recipe/category scanning via bucketed profession-window events.
Additional hardening and correctness updates:

Adds nil-safe guards around Blizzard tradeskill API reads.
Fixes tradeskill report formatting to use skillMod consistently.
Makes tradeskill table updates safer by deferring removals until after iteration.
Improves recipe-link handling when links are missing or malformed.
Follow-up cleanup in this commit:

Adds debug trace logs for both refresh paths to make in-game verification easier.
Wraps bucketed trade-window scanning in a named helper callback for clearer instrumentation.
Moves the helper to file scope for cleaner OnEnable registration flow and readability.
Behavioral intent:

Keep fast no-window profession level updates.
Keep detailed window-based recipe/category persistence.
Use both methods without duplicate or noisy event handling paths.
Optional testing notes to include:

Verify SKILL_LINES_CHANGED and LEARNED_SPELL_IN_SKILL_LINE trigger debounced current-profession refresh.
Verify opening/updating profession UI triggers bucketed detailed ScanTrade path.
Verify rapid skill-line bursts collapse into a single refresh cycle.
Verify tradeskill report displays modifier values from skillMod.

### DIFF
--- a/WoWPro/WoWPro.lua
+++ b/WoWPro/WoWPro.lua
@@ -110,7 +110,12 @@ WoWPro:Export("Error")
 
 local function ends_with(str, ending)
     return ending == "" or str:sub(-#ending) == ending
- end
+end
+
+local function QueueTradeWindowScan(...)
+    WoWPro:dbp("QueueTradeWindowScan() running ScanTrade()")
+    WoWPro.ScanTrade(...)
+end
 
 -- WoWPro Log function all only --
 function WoWPro.LogCall(nomen, ...)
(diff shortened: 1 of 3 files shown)


Ludovicus-Maior:
Though short, this it an interesting approach to the skill line issue.
I would also note that TSM has recently acquire the capacity to scan trade-skills without opening profession windows.
I need to see what they are doing to see if we can integrate a similar skill into WoWPro.
We just need the current professions and skills, so it should be a much smaller chunk of code.

Ludovicus-Maior:
@Spoony36 Dude!  Interesting approach.  It does not conform with the way TSM does it, as it opens a profession window and then takes a couple of seconds to refresh the information, which implies it did some query and then got information.  I think we need both methods to make sure we have the right information.

Ludovicus-Maior (approved):
It looks pretty good, but I'll need to run it for a bit and make sure it behaves.

Ludovicus-Maior:
@Spoony36 So far, I have not noticed anything wrong in retail.
Perhaps @Hendo72 can give it a good run in Classic.

PR #3435, opened 2026-08-04
reduce guide update churn from loot autocomplete events

Spoony36:
AutoComplete loot processing was triggering guide refreshes on every bucketed loot/bag event, even when no step state changed. During crafting/loot bursts this could cause repeated UpdateGuideReal passes and visible stutter for some users.

Changes:

Updated AutoCompleteLoot to track whether guide state actually changed.
Set a local flag only when a loot step is newly completed.
Call UpdateGuide("WoWPro.AutoCompleteLoot") only when that flag is true.
Preserve existing completion behavior and tracking text updates.

Impact:
Significantly fewer unnecessary guide recomputations during CHAT_MSG_LOOT/BAG_UPDATE bursts.
Lower UI work during high-frequency loot activity.
No functional change to loot-step completion semantics.

### DIFF
--- a/WoWPro/WoWPro_AutoComplete.lua
+++ b/WoWPro/WoWPro_AutoComplete.lua
@@ -250,6 +250,7 @@ end
 -- Auto-Complete: Loot based --
 function WoWPro.AutoCompleteLoot()
     if not WoWPro.GuideLoaded then return end
+    local guideStateChanged = false
     for i = 1, 1 + WoWPro:GetActiveStickyCount() do
         local index = WoWPro.rows[i].index
         local lootItems = WoWPro.lootitem[index]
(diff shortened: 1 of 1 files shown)


Ludovicus-Maior:
Hmmm!  Does this mean the counts of the loot would not be updated as you looted the items?

Spoony36:
> Hmmm! Does this mean the counts of the loot would not be updated as you looted the items?

No, the loot counts will still update as items are looted. AutoCompleteLoot now refreshes each affected row through UpdateQuestTrackerRow, which updates the displayed count and preserves GroupSync broadcasts without rebuilding the entire guide.

When collecting the required quantity completes a step, CompleteStep already triggers the necessary guide refresh. I’ve updated the PR accordingly and removed the redundant extra refresh.

Ludovicus-Maior:
Thanks for the information.

PR #3436, opened 2026-08-05
Main frame refactor 2.0 (guide window)

Hendo72:
Refactor to enforce expansionAnchor as the authoritative anchor for all WoWPro
frame positioning and layout behavior:

- Removed legacy padding-based layout system and all geometry-derived anchor
  recalculations that bypassed or overrode expansionAnchor.
- Consolidated all frame positioning logic so expansionAnchor is the single
  source of truth for MainFrame placement and expansion direction.
- Added AnchorOffset() to compute vertical stack height strictly from
  expansionAnchor and visible UI components.
- Introduced Reanchor() to apply offsets, reposition GuideFrame, and recalc
  layout without inferring anchors from frame geometry.
- Updated AnchorRestore() to restore expansionAnchor, size, and scale, then
  reapply layout via Reanchor() + AnchorOffset().
- Standardized drag pipeline so expansionAnchor is only updated on mouse-up,
  eliminating implicit anchor rewrites during movement.
- Implemented proxy drag forwarding from Titlebar and ButtonBar to MainFrame
  to ensure consistent anchor updates.
- Converted ButtonBar to a Button frame and guarded click/drag registration to
  prevent nil-method calls.
- Removed all fallback anchors and deprecated code paths that attempted to
  derive anchors from frame position or size.

** _I strongly feel that this is evolving into a 4th module and is becoming the WoWPro_UI._

### DIFF
--- a/WoWPro/WoWPro_CompatUtil.lua
+++ b/WoWPro/WoWPro_CompatUtil.lua
@@ -219,13 +219,21 @@ function WoWPro.SuperTrack_SetSuperTrackedQuestID(questID)
 end
 
 function WoWPro.SetResizeBounds(frame, minWidth, minHeight, maxWidth, maxHeight)
+    -- Safe defaults
+    minWidth  = minWidth  or 100
+    minHeight = minHeight or 40
+
+    -- Clamp max size to screen if not provided
+    maxWidth  = maxWidth  or UIParent:GetWidth()
+    maxHeight = maxHeight or UIParent:GetHeight()
+
     if frame.SetResizeBounds then
-       frame:SetResizeBounds(minWidth, minHeight, maxWidth, maxHeight)
+        -- Retail 10.0+
+        frame:SetResizeBounds(minWidth, minHeight, maxWidth, maxHeight)
     else
+        -- Classic fallback
         frame:SetMinResize(minWidth, minHeight)
-        if maxWidth then
-            frame:SetMaxResize(maxWidth, maxHeight)
-        end
+        frame:SetMaxResize(maxWidth, maxHeight)
     end
 end
 
--- a/WoWPro/WoWPro_Frames.lua
+++ b/WoWPro/WoWPro_Frames.lua
@@ -217,27 +217,27 @@ function WoWPro:PaddingSet()
     local pad = WoWProDB.profile.pad
     -- Padding Customization --
     if WoWPro.Titlebar:IsShown() then
-        WoWPro.StickyFrame:SetPoint("TOPLEFT", WoWPro.Titlebar, "BOTTOMLEFT", pad+3, -pad+3)
-        WoWPro.StickyFrame:SetPoint("TOPRIGHT", WoWPro.Titlebar, "BOTTOMRIGHT", -pad-3, -pad+3)
+        WoWPro.StickyHeader:SetPoint("TOPLEFT", WoWPro.Titlebar, "BOTTOMLEFT", pad+3, -pad+3)
+        WoWPro.StickyHeader:SetPoint("TOPRIGHT", WoWPro.Titlebar, "BOTTOMRIGHT", -pad-3, -pad+3)
     else
         -- Match the same vertical padding as the titlebar-shown case
-        WoWPro.StickyFrame:SetPoint("TOPLEFT", pad+3, -pad+3)
-        WoWPro.StickyFrame:SetPoint("TOPRIGHT", -pad-3, -pad+3)
+        WoWPro.StickyHeader:SetPoint("TOPLEFT", pad+3, -pad+3)
+        WoWPro.StickyHeader:SetPoint("TOPRIGHT", -pad-3, -pad+3)
     end
-    WoWPro.GuideFrame:SetPoint("TOPLEFT", WoWPro.StickyFrame, "BOTTOMLEFT" )
-    WoWPro.GuideFrame:SetPoint("TOPRIGHT", WoWPro.StickyFrame, "BOTTOMRIGHT" )
+    WoWPro.GuideFrame:SetPoint("TOPLEFT", WoWPro.StickyHeader, "BOTTOMLEFT" )
+    WoWPro.GuideFrame:SetPoint("TOPRIGHT", WoWPro.StickyHeader, "BOTTOMRIGHT" )
     -- Only anchor to bottom when scrolling is enabled to constrain height
     if WoWProDB.profile.guidescroll then
         WoWPro.GuideFrame:SetPoint("BOTTOM", 0, pad)
-        local stickyHeight = WoWPro.StickyFrame:IsShown() and WoWPro.StickyFrame:GetHeight() or 0
+        local stickyHeight = WoWPro.StickyHeader:IsShown() and WoWPro.StickyHeader:GetHeight() or 0
         local mainHeight = WoWPro.MainFrame:GetHeight() or 0
         local guideHeight = math.max(mainHeight - stickyHeight - (pad * 2), 25)
         WoWPro.GuideFrame:SetHeight(guideHeight)
     else
         WoWPro.GuideFrame:ClearAllPoints()
-        WoWPro.GuideFrame:SetPoint("TOPLEFT", WoWPro.StickyFrame, "BOTTOMLEFT" )
-        WoWPro.GuideFrame:SetPoint("TOPRIGHT", WoWPro.StickyFrame, "BOTTOMRIGHT" )
-        local stickyHeight = WoWPro.StickyFrame:IsShown() and WoWPro.StickyFrame:GetHeight() or 0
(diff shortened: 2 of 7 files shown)


Spoony36:
Here's my review, might want to get @Ludovicus-Maior & @Furydins opinion as to whether they agree or disagree. Might also be intended behaviour

Blocking: Position migration is missing (users can lose saved frame location)
Where to place:
[WoWPro_Frames.lua:1120]
Comment text:
This switches anchor persistence from profile.position to profile.anchorX/profile.anchorY, but I do not see a one-time migration path. Existing users with only profile.position will restore to default offsets (0,0) for their expansion anchor after update. This is a user-visible regression.

Suggested fix:
Add a one-time migration before first restore, converting legacy profile.position into anchor fields (including pct-mode if present), then keep using the new model.

Suggested snippet:
if not WoWProDB.profile.anchorX and type(WoWProDB.profile.position) == "table" then
local pos = WoWProDB.profile.position
local anchor = pos[1] or WoWProDB.profile.expansionAnchor or "TOPLEFT"
WoWProDB.profile.expansionAnchor = anchor

if pos[6] == "pct" and pos[7] and pos[8] then
local uiW = UIParent:GetWidth() or 0
local uiH = UIParent:GetHeight() or 0
WoWProDB.profile.anchorX = (pos[7] or 0) * uiW
WoWProDB.profile.anchorY = (pos[8] or 0) * uiH
else
WoWProDB.profile.anchorX = pos[4] or 0
WoWProDB.profile.anchorY = pos[5] or 0
end
end

High: Drag lock option appears bypassed
Where to place:
[WoWPro_Config.lua:244]
Comment text:
The drag toggle is user-facing config, but this refactor moves drag handling into multiple frame scripts and no longer consistently gates movement on profile.drag. If this lands as-is, users can still move the window even when drag is disabled.

Suggested fix:
Use one shared guard in every drag entry point:
if not WoWProDB.profile.drag then return end

High: Mouseover-notes behavior looks unintentionally removed
Where to place:
[WoWPro_Frames.lua:603]
Comment text:
Current behavior conditionally hides inline note text and shows mouseover notes when profile.mousenotes is enabled. The refactor’s RowSizeSet simplification removes this branch, so the mousenotes setting likely becomes ineffective/regressed.

Suggested fix:
Keep the mousenotes branch in row layout or move it to a dedicated helper that RowSizeSet still calls.

High: New setting key does not match existing schema
Where to place:
[WoWPro_Frames.lua:395]
Comment text:
I see writes to showLeftSideButtons in the refactor, but the addon consistently uses profile.leftside for side placement state. This looks like a key mismatch and would silently break side-switch correction logic.

Suggested fix:
Replace showLeftSideButtons with leftside, unless a full schema migration/default wiring is added.

Medium: On-screen clamp safeguards were removed
Where to place:
[WoWPro_Frames.lua:297]
Comment text:
This removes ClampBarsOnScreen and related overflow correction logic. Even if anchor unification is the goal, this increases risk of bar/button clipping at screen edges, especially at non-1.0 UI scales.

Suggested fix:
Keep a minimal post-action clamp pass after drag, resize, and restore.

Medium: Indirect callback invocation is brittle
Where to place:
[WoWPro_Frames.lua:1567]
Comment text:
Calling MainFrame:GetScript("OnMouseUp")(...) assumes that callback always exists and remains compatible. This can hard-error if setup order changes or script gets replaced.

Suggested fix:
Extract shared save/reanchor logic into a normal function and call it directly from all mouse-up handlers. At minimum, nil-guard before invoking fetched scripts.

Medium: Removed cropping warning changes UX without replacement
Where to place:
[WoWPro_Frames.lua:603]
Comment text:
The previous CroppedGuideWarning flow informed users when screen height constrained visible rows. This is removed, so users may see truncated guides with no explanation.

Suggested fix:
Either keep a lightweight warning path or add equivalent feedback in the new layout model.

WoWPro/WoWPro_CompatUtil.lua
```diff
@@ -219,13 +219,21 @@ function WoWPro.SuperTrack_SetSuperTrackedQuestID(questID)
 end
 
 function WoWPro.SetResizeBounds(frame, minWidth, minHeight, maxWidth, maxHeight)
+    -- Safe defaults
+    minWidth  = minWidth  or 100
+    minHeight = minHeight or 40
+
+    -- Clamp max size to screen if not provided
+    maxWidth  = maxWidth  or UIParent:GetWidth()
```
Spoony36, line 227: Would this be better?
    local ui = _G.UIParent
    maxWidth  = maxWidth  or (ui and ui:GetWidth() or 0)
    maxHeight = maxHeight or (ui and ui:GetHeight() or 0)

WoWPro/WoWPro_CompatUtil.lua
```diff
@@ -219,13 +219,21 @@ function WoWPro.SuperTrack_SetSuperTrackedQuestID(questID)
...
-       frame:SetResizeBounds(minWidth, minHeight, maxWidth, maxHeight)
+        -- Retail 10.0+
+        frame:SetResizeBounds(minWidth, minHeight, maxWidth, maxHeight)
     else
+        -- Classic fallback
         frame:SetMinResize(minWidth, minHeight)
-        if maxWidth then
-            frame:SetMaxResize(maxWidth, maxHeight)
-        end
+        frame:SetMaxResize(maxWidth, maxHeight)
```
Spoony36, line 236: if maxWidth > 0 and maxHeight > 0 then
            frame:SetMaxResize(maxWidth, maxHeight)
        end

?????

WoWPro/WoWPro_Frames.lua
```diff
@@ -1843,14 +1329,31 @@ function WoWPro:CreateTitleBar()
...
+
+        WoWPro.InhibitAnchorStore   = false
+        WoWPro.InhibitAnchorRestore = false
+        WoWPro.InhibitReanchor      = false
+    end)
+    WoWPro.Titlebar:SetScript("OnMouseUp", function(self, button)
+        if button ~= "LeftButton" then return end
+        if InCombatLockdown() then return end
+
+        WoWPro.MainFrame:GetScript("OnMouseUp")(WoWPro.MainFrame, "LeftButton")
```
Spoony36, line 1356: local onMouseUp = WoWPro.MainFrame and WoWPro.MainFrame:GetScript("OnMouseUp")
        if onMouseUp then
            onMouseUp(WoWPro.MainFrame, "LeftButton")
        end

WoWPro/WoWPro_Frames.lua
```diff
@@ -1565,18 +1019,49 @@ end
...
+
+        WoWPro.InhibitAnchorStore   = false
+        WoWPro.InhibitAnchorRestore = false
+        WoWPro.InhibitReanchor      = false
+    end)
+    WoWPro.ButtonBar:SetScript("OnMouseUp", function(self, button)
+        if button ~= "LeftButton" then return end
+        if InCombatLockdown() then return end
+
+        WoWPro.MainFrame:GetScript("OnMouseUp")(WoWPro.MainFrame, "LeftButton")
```
Spoony36, line 1063: local onMouseUp = WoWPro.MainFrame and WoWPro.MainFrame:GetScript("OnMouseUp")
        if onMouseUp then
            onMouseUp(WoWPro.MainFrame, "LeftButton")
        end

WoWPro/WoWPro_NilGuide.lua
```diff
@@ -8,8 +8,8 @@ function WoWPro:LoadNilGuide()
     WoWPro.stepcount = 1
 
     WoWPro.TitleText:SetText("No Guide Loaded")
-    WoWPro.StickyFrame:Hide()
-    WoWPro.StickyFrame:SetHeight(1)
+    WoWPro.StickyHeader:Hide()
```
Spoony36, line 11: local sticky = WoWPro.StickyHeader or WoWPro.StickyFrame
    if sticky then
        sticky:Hide()
        sticky:SetHeight(1)
    end

WoWPro/WoWPro_NilGuide.lua
Spoony36, line 36: if WoWPro.PaddingSet then
        WoWPro.PaddingSet()
    end

Spoony36 (commented):
Just a few thoughts, might not be the way forward but offering suggestions

Furydins:
On the issue of losing existing positioning - migrating the settings is nice for users - but I'd not see it as an absolute blocker if it's not easy to do.

We can always just call out that there will be a one time reset to make the positioning work better - I suspect most folks are OK with having to correct position as long as it's a one off.

The downside of having migration code is what happens to it post migration? - we've no way of knowing how long folks go between updates - so it's hard to remove it - meaning it tends to hang around like a bad smell till it bites you in the behind years down the line!

Spoony36:
image

we seem to have lost step text too at the moment but it's only draft so something we can work on

Hendo72:
This PR is faulty and has been scrapped for a newer version

PR #3444, opened 2026-08-15
No more peeking at auras in combat

Furydins:
Addresses user reported issue

- The 12.1.0 restrictions mean we're no longer able to interact with aura data at all in combat so I've added strict combat checks for all calls.
- This will likely delay step updates/skips till we exit combat - not ideal but we are where we are.

Tested with the Maldraxxus guide which uses |BUFF| a fair bit and it seems to behave sensibly - but me may have to watch for corner cases.

### DIFF
--- a/WoWPro/WoWPro_Broker.lua
+++ b/WoWPro/WoWPro_Broker.lua
@@ -3886,6 +3886,10 @@ function WoWPro.NextStep(guideIndex, rowIndex)
             -- This tests for spells that are cast on you and show up as buffs
             if WoWPro.buff and WoWPro.buff[guideIndex] then
                 local buff = WoWPro.buff[guideIndex]
+                if _G.InCombatLockdown() then
+                    WoWPro:dbp("Skipping buff check because in combat.", buff)
+                    break
+                end
                 local buffy = WoWPro:CheckPlayerForBuffs(buff)
                 if buffy then
                     skip = true
--- a/WoWPro/WoWPro_Events.lua
+++ b/WoWPro/WoWPro_Events.lua
@@ -226,7 +226,7 @@ end
 
 
 WoWPro.RegisterEventHandler("UNIT_AURA", function(event, ...)
-    if not WoWPro.MaybeCombatLockdown() then
+    if not _G.InCombatLockdown() then
         WoWPro.AutoCompleteBuff(...)
     end
     end)


Spoony36:
Do we need to look at these?

2. Target GUID Guard Runs Too Late
Priority: High
Impact: Target-related completion may error or stop working during combat.

WoWPro.lua:776 currently performs a boolean operation before checking secrecy:


If UnitGUID("target") is secret, the or "" operation itself may be prohibited. The later check cannot protect it.

Recommended action:

Store the raw result first, check issecretvalue(), and only then handle nil, split it, compare it, or log it.


Acceptance test:

Change targets repeatedly during restricted combat and verify that target/NPC guide steps produce no Lua errors. Confirm target completion resumes when readable data becomes available.

3. Group Sync Can Be Blocked in Combat
Priority: Medium
Impact: Party progress and tracking messages may be silently lost.

Unconditional party addon messages are sent from:

WoWPro.lua:938
WoWPro_Broker.lua:943
WoWPro_Broker.lua:1459
WoWPro_Broker.lua:2078
WoWPro_Events.lua:484
Midnight added Enum.SendAddonMessageResult.AddOnMessageLockdown. The return value from C_ChatInfo.SendAddonMessage() is currently ignored.

Recommended action:

Route all Group Sync sends through one helper.
Inspect the send result.
When the result is AddOnMessageLockdown, retain the latest group, step, and tracking state.
Flush deferred state after PLAYER_REGEN_ENABLED or ADDON_RESTRICTION_STATE_CHANGED.
Coalesce repeated step/track messages rather than queueing every update.
Acceptance test:

Enable Group Sync, enter a restricted encounter, change guide progress, and verify that the other party member receives the final state after restrictions end.

4. Other Secret Guards Use Unsafe Ordering
Priority: Medium/Low
Impact: Situational errors when Blizzard returns a secret location value.

Affected hearth-location paths include:

WoWPro_Events.lua:249
WoWPro_Events.lua:865
WoWPro_AutoComplete.lua:419
Patterns such as this compare the value before checking secrecy:


The secret check must come first.

Recommended action:

Introduce a small helper for readable values, or consistently use this order:


Also audit developer commands that directly format aura names, spell IDs, or coordinates, particularly WoWPro_Command.lua:108 and the coordinate displays in WoWPro.lua:712.

Existing Protections That Look Sound
Secure button attributes are generally configured outside combat.
Override binding changes are guarded by InCombatLockdown().
Protected frame positioning and resizing have guards.
WoWPro does not register COMBAT_LOG_EVENT or COMBAT_LOG_EVENT_UNFILTERED.
Midnight target macros already omit the raid-marker Lua snippet.
The TOC declares current 12.0 interfaces in WoWPro.toc:1.
Suggested Work Order
Fix secret-value ordering in aura and target GUID paths.
Stop unconditionally forcing RecklessCombat = true.
Add a centralized, deferred Group Sync sender.
Audit remaining API returns before they are compared, logged, formatted, or used as table keys.
Test inside a dungeon or raid where Blizzard’s restricted combat state is active, not only against an outdoor training dummy.

Furydins:
Great suggestions and we probably should look at them but it feels like too much scope creep for a bugfix PR!

I'd suggest capturing them somewhere as discrete tasks  we can address separately.

If there's something I missed in this that is related to the issue Doublebass is having do call it out!

Ludovicus-Maior:
Sigh

PR #3453, opened 2026-09-12
Guide window jumps off the bottom of the screen on update — UIParent and frame coordinate spaces are mixed

nitfol:
### The problem

In the Classic client, on both my Windows PC and my macOS PC, the WoWPro window jumps to a
position off the bottom of the screen any time it is updated. It seems to think it is not
fully visible — which it is — and tries to relocate itself fully on screen, but miscalculates
and pushes itself off the bottom instead.

It ultimately appears to be a problem with scaled coordinates in the newer WoW client.

`GetLeft()` / `GetRight()` / `GetTop()` / `GetBottom()` report coordinates in **each frame's
own scaled space**, and `SetPoint` offsets are interpreted in the **anchored frame's** space
— while `UIParent`'s edges and the screen size derived from them are in **UIParent's** space.

The anchor and clamp code in `WoWPro_Frames.lua` compares the two directly. That is exact
when `MainFrame:GetEffectiveScale() == UIParent:GetEffectiveScale()` and wrong by their ratio
otherwise, so it behaves at UI scale 1.0 and degrades progressively away from it.

Separately, the screen-size fallback derives height from physical pixels, which overstates it
by `physH / 768`: WoW's UI coordinate space is always 768 units tall regardless of
resolution. That is the one that pushes frames off the bottom edge.

### The change

Three helpers, then the call sites go through them:

- **`GetUIScreenSize()`** — the screen in *UIParent* coordinates, which is the space every
  cross-space comparison in this file ultimately resolves against. The fallback goes through
  the fixed 768-unit height rather than raw pixels.
- **`UIParentRatio(frame)`** — `frame:GetEffectiveScale() / UIParent:GetEffectiveScale()`,
  applied wherever a frame-space coordinate meets a UIParent-space one.
- **`GetCornerOffsets(corner)`** — the `SetPoint` offsets, in MainFrame's own space, that
  leave the frame exactly where it already is when re-anchored to `corner` of UIParent. Also
  returns them in UIParent space, which is what gets persisted, so a stored position survives
  a later scale change.

Updated: `ResetMainFramePosition`, `SetExpansionAnchor`, `AnchorStore`, `AnchorRestore`,
`ClampBarsOnScreen`, `ClampSideButtonsOnScreen`, `SetDynamicResizeBounds`,
`AdjustMainFrameForVisibleButtonBar`, `AdjustMainFrameForHiddenButtonBar`,
`ContractGuideToRows`, `RowSizeSet`, `SetMouseNotesPoints`, `DisableLeftHandedIfOffScreen`.

Net `+160 / −177` in one file.

### Invariants it is meant to hold, at any scale

- `ResetMainFramePosition()` and `SetExpansionAnchor(corner)` never *move* the frame.
- `ClampBarsOnScreen()` leaves an on-screen bar alone, and pulls an overflowing one to
  exactly 2 units inside the top edge.
- `AnchorStore` / `AnchorRestore` round-trip.

### Testing

- **Verified by use.** It has been running on my install since 2026-08-14, roughly four
  weeks, on macOS 26.5.1 / MacBook Air M2 / built-in Liquid Retina 2880x1864, at WoW's
  **default** UI scale. Client 1.15.9 (Classic Era), WoWPro 2026.08.27.A.
- **Passes `luacheck`.** `luacheck . --no-color -q` reports 0 warnings / 0 errors across
  1317 files with this patch applied, exactly matching the unpatched baseline. Re-checked
  today with luacheck 1.2.0 against the repo's own `.luacheckrc`.


rmburgess-dev:
Your analysis is right and I have merged this into #3468 so it can go in with the rest of the Forever work.

I confirmed the arithmetic independently before merging: `GetUIScreenSize()` divides physical pixels by the effective scale, and WoW's UI space is always 768 units tall. The error is not small —

```
resolution   uiScale   computed   correct   error
1080p        1.0000     1080.0     768       +41%
1080p        0.7111     1518.8     768       +98%
1440p        1.0000     1440.0     768       +88%
2160p        1.0000     2160.0     768      +181%
```

It only agrees at a scale that happens to equal `physH / 768`, which is why it behaves at some settings and runs the window off the bottom at others. Your `physH / 768` fallback and the UIParent-space helpers are the correct fix, and the frame-space-versus-UIParent-space distinction is the right way to describe it.

**One thing to know before it lands: the branch is five commits behind `main`** (based on `f7c84c7`, the same base as #3456). Merged as-is it deletes the button-config work from #3452 — all five `show*Button` defaults, the `Button Display Options` group and the guards in `WoWPro_Broker.lua`. Merging it onto current `main` resolves that cleanly, which is what #3468 does; if you would rather rebase it yourself and land it separately, that is fine too and I will drop it from mine.

Also worth flagging: this is the second PR in a row with the same stale-base problem, so it may be worth a rebase before merging anything currently open.

Verified after merging: 1017 .lua files pass `luac -p`, `validate_tocs.py` reports 35 TOCs and 1772 referenced files all resolving, and the button config from #3452 is intact. Thanks for the write-up — the "screen size in UIParent coordinates" framing is what made it quick to check.

PR #3455, opened 2026-09-16
Refactor row update

Hendo72:
Extract RowUpdate helpers from Broker and isolate row-menu behavior

### Summary:
- Moves RowUpdate-specific logic out of Broker and into a dedicated RowUpdate file.
- Creates self-contained helpers for row rendering, dropdown generation, and related row behavior so Broker is no longer the owner of the update logic.
- Keeps Broker responsible for global lifecycle/state initialization while the extracted helpers safely guard against nil table cases and legacy call paths.
- Improves maintainability by reducing Broker coupling and keeping row update behavior grouped in one place.


Hendo72:
@copilot resolve the merge conflicts in this pull request

Ludovicus-Maior (approved):
This is what I expected to see.

PR #3456, opened 2026-09-17
Wow: Forever support

Furydins:
- Add ToCs for Forever 
- The devs will regret using their internal codename in production filenames (camelot) ;-)
- Address errors due to Trade skills APIs returning invalid responses (possibly disabled on Beta?)
- Create new WoWPro.FOREVER flag and configure
- Clone Hendo's classic guides as placeholder
- Configure TOC to load cloned guides
-  Disable all modules except Levelling by default

TODO: Verify professions once the issues with tradeskills APIs are resolved by Blizzard
TODO: Figure out what I've forgotten to update

Notes: 

The majority of "changed" files are the guide clones-  the actual code changes are fairly limited.
Saved variables are not being loaded after login or /reload - which musses up skipped steps and settings persistence - This is confirmed as a Blizzard side issue.


rmburgess-dev:
Thanks for doing this — the `WoWPro.FOREVER` flag, the Forever zone data and the tradeskill guards are the right kind of groundwork. I've been setting up the same thing locally and hit a few concrete problems in this branch that are worth fixing before it merges. All of it is reproducible from the branch as pushed.

## 1. The branch is stale and would revert merged work

`f7c84c7` is the merge base; `main` has moved 5 commits since, including #3452 "Added configuration to enable/disable extra buttons independently". Because this branch predates it, merging as-is **deletes** those options:

```
 WoWPro/WoWPro.lua     | 5 -----      showItemButton, showTargetButton, showEAButton,
                                      showJumpButton, showLootsButtons
 WoWPro/WoWPro_Broker.lua  | removes the guards that read them
 WoWPro/WoWPro_Config.lua  | removes the "Button Display Options" config group
```

A rebase onto current `main` should resolve all of it without conflict, since the two changes touch different concerns.

## 2. Three of the Camelot TOCs don't declare interface 16001

| TOC | `## Interface:` |
| --- | --- |
| `WoWPro_Camelot.toc` | `16001` ✅ |
| `WoWPro_Profession_Camelot.toc` | `16001` ✅ |
| `WoWPro_WorldEvents_Camelot.toc` | `16001` ✅ |
| `WoWPro_Leveling_Camelot.toc` | `120007,120100` ❌ |
| `WoWPro_Achievements_Camelot.toc` | `120007,120100` ❌ |
| `WoWPro_Recorder_Camelot.toc` | `120007,120100` ❌ |
| `WoWPro_Dailies_Camelot.toc` | `11509` ❌ |

The whole point of a `_Camelot.toc` is to be the file Forever selects, so it should carry Forever's interface version. Carrying `120007,120100` (Retail) makes the client treat those three as out of date, and `11509` does the same for Dailies. Leveling is the one users actually look at, so it's the one that matters most.

## 3. `WoWPro_Leveling/Forever/` duplicates Vanilla byte-for-byte

The PR body says "Clone Hendo's classic guides as placeholder", and that is literally what happened: all **63** `.lua` files under `WoWPro_Leveling/Forever/` are byte-identical to the existing `WoWPro_Leveling/Vanilla/` files — 1,303,792 bytes, 11,595 of the PR's 11,736 added lines:

```
WoWPro_Leveling/Forever/Alliance/01_12_Elwynn.lua
  == WoWPro_Leveling/Vanilla/Alliance/01_12_Elwynn.lua   (md5 identical)
... 63/63 files, 0 genuinely changed
```

It gets worse than wasted bytes: because the copies are byte-identical they also register the **same guide IDs** (`WoWPro:RegisterGuide('ClassicElwynn0112', ...)`). Any client that ends up loading both the `Vanilla` and the `Forever` XML — a second TOC pointing at `Vanilla\`, a user's own tweak, or a future `_Classic.toc` fallback — will silently have one set overwrite the other, with no error.

Since the content is identical, `WoWPro_Leveling_Camelot.toc` can just point at the existing Vanilla guides:

```
Vanilla\Alliance\Guides.xml
Vanilla\Neutral\Guides.xml
Vanilla\Horde\Guides.xml
```

That drops ~11.6k lines from the PR and keeps a single copy of the guides to maintain. If Forever genuinely diverges later, the clones can come back — but they should be actual edits, not copies, or the guide IDs need distinct prefixes.

## 4. Two TOCs pull in Retail content

`WoWPro_Profession_Camelot.toc` loads:

```
Retail\Guides.xml
Vendors\Guides.xml
Vanilla\Guides.xml
```

`Retail\Guides.xml` brings in the Retail profession guides (`Alchemy_BfA.lua`, and so on) on a level 60 client. `WoWPro_Profession_Vanilla.toc` loads `Vanilla\Guides.xml` only.

`WoWPro_WorldEvents_Camelot.toc` loads `Retail\{Alliance,Neutral,Horde}\Guides.xml`. There is a `WoWPro_WorldEvents\Vanilla\{Alliance,Horde,Neutral}\Guides.xml` that lists the holiday guides, which is what Vintage content should get. (Note `WoWPro_WorldEvents_Vanilla.toc` references no guide XML at all, so that Vanilla data is currently unreachable on every client — probably its own bug.)

## 5. `GetNumSkillLines` / `GetSkillLineInfo` exist on Forever

Minor, but the new guard is dead code. Both globals are present in the Forever client binary:

```
GetNumSkillLines    2 occurrences
GetSkillLineInfo    8 occurrences
C_Professions       0 occurrences   <- does not exist
```

So `if not _G.GetNumSkillLines then ... end` never fires, and `C_Professions` is not the modern replacement to move to. If tradeskills are returning bad data on the beta, the guard isn't what's covering it.

## Tooling

I wrote a small validator that checks every path a `.toc` references actually exists, following XML `<Script>`/`<Include>` chains and ignoring commented-out entries, and that a `.toc` name uses a suffix the client actually searches for. It's proposed separately in #3462. On this branch it reports clean — `Checked 35 .toc files, 1867 referenced files. OK` — so none of the above is a missing-file problem; it's all version and content selection.

rmburgess-dev:
**Correction to point 5 of my earlier review.** I said the `GetNumSkillLines` guard is dead code because the global is present in the client binary. That was wrong, and I've now seen the runtime evidence:

```
Interface/AddOns/WoWPro/WoWPro_Trade.lua:382: attempt to call a nil value
[C]: in function 'GetNumSkillLines'
[Interface/AddOns/WoWPro/WoWPro_Trade.lua]:382: in function 'UpdateTradeSkills'
[Interface/AddOns/WoWPro/WoWPro_Events.lua]:277: in function '?'
```

`GetNumSkillLines` and `GetSkillLineInfo` do appear in the Forever client binary, which is what I based the claim on — but only as members of the `C_SkillInfo` namespace:

```
C_SkillInfo.AbandonSkill / CollapseSkillHeader / ExpandSkillHeader / GetNumSkillLines /
GetSelectedSkill / GetSkillLineInfo / GetSkillLineInfoByID / SetSelectedSkill
Usage: local skillLineAttributes = C_SkillInfo.GetSkillLineInfo(index)
Usage: local skillLineAttributes = C_SkillInfo.GetSkillLineInfoByID(ID)
```

The bare globals are not callable, so your guard fires exactly as intended and is doing real work. Apologies for the noise — searching the binary for an identifier is not evidence that the API exists at runtime, which I should have allowed for before writing it up as dead code.

The other four points stand. Fix for this one, plus the earlier `C_Spell.GetSpellInfo` nil crash, is in #3464.

rmburgess-dev:
## Verified positive: the Camelot TOC approach works

I've now driven this end to end on a live Forever beta client (1.60.1, interface 16001). Adding a `_Camelot.toc` is the right mechanism and the client does select it — I confirmed against the SavedVariables the client wrote:

```lua
["currentguide"] = "ClassicTirisfalGlades0112",
```

A Vanilla guide ID, with the addon's internal log showing 11 quests populated and live objective updates arriving:

```
Quest Log populated. 11 quests found.
Updated objective #1 (3/7 Putrid Claw) on quest [A Putrid Task]
UpdateGuideReal(77): ActiveStep=A A Rogue's Deal¦QID¦590¦PRE¦8¦M¦38.26,56.79¦Z¦1420...
```

215 log lines, no error or warning entries. So the core of this PR is confirmed working on the real client, which is more than I could say when I wrote the review above.

Two follow-ups to the review, both now with runtime evidence rather than inference:

**The `GetNumSkillLines` guard is load-bearing, and I was wrong to call it dead code.** Runtime on Forever:

```
Interface/AddOns/WoWPro/WoWPro_Trade.lua:382: attempt to call a nil value
[C]: in function 'GetNumSkillLines'
```

The globals are not callable on Forever; only `C_SkillInfo.*` exists. Your guard is correct and necessary — please keep it through the rebase. Fix for the neighbouring `C_Spell.GetSpellInfo` nil crash is in #3464.

**`WoWPro.CLASSIC` is now `>= 10000 and < 16000`, and five other call sites still read it as "is this a Classic client".** That is the right direction, but Forever is a third thing: a Classic-content client with a modern API surface. The live failure above is the proof — `CLASSIC` was true, so the code took the Classic skill-line path, and that path does not exist there. Worth checking each remaining `WoWPro.CLASSIC` site for whether it means "Classic content" or "Classic API":

```
WoWPro/WoWPro_Trade.lua:9        profession tables      <- already handled by this PR
WoWPro_Leveling/WoWPro_Leveling.lua:129   starter guides  <- already handled by this PR
WoWPro/WoWPro_Broker.lua:3493, 3749       SoD rune spells, also gated on C_Seasons
WoWPro/WoWPro_Events.lua:287              NEUTRAL_FACTION_SELECT_RESULT
```

The two Broker sites are harmless on Forever because `C_Seasons` is nil there, but they read as accidents rather than decisions.

rmburgess-dev:
**One more correction, and it is in your favour.** I implied in the review that `WoWPro_ZoneData_Forever.lua` was mostly a re-statement of what `main` already has. That was based on counting `DefineZone1` calls without comparing their contents, which was sloppy. Doing it properly:

```
ZoneData_Forever DefineZone1 ids: 59
main ZoneData.lua DefineZone1 ids: 54
in main but NOT in Forever: (none)
in Forever but not main: 2482, 2521, 2524, 2548, 2652
```

Your file is a strict superset, and the five extra entries are Forever's own world:

```lua
DefineZone1(2482, "Mount Hyjal",       UIMapType.Zone, 1414, nil)
DefineZone1(2521, "Zephras Isle",      UIMapType.Zone,  947, nil)
DefineZone1(2524, "Darkspear Islands", UIMapType.Zone,  947, nil)
DefineZone1(2548, "Riverglades",       UIMapType.Zone, 1415, nil)
DefineZone1(2652, "Shen'dralas",       UIMapType.Zone, 1414, nil)
```

Zephras Isle, Riverglades and Shen'dralas are new Forever zones that no existing zone table covers, so this file is doing real work and is needed for the addon to place guides on those maps at all. It also extends the child lists on the existing continents. Please keep it.

Re-defining on top of `WoWPro_ZoneData.lua` is safe, for anyone else reading: `DefineZone1` assigns `WoWPro.MapInfo[mapId]` and `WoWPro.Zone2MapID[zone]` outright, so last write wins, and both files run under `if WoWPro.Client ~= 1 then return end`. The only cost is one `DupCheck` debug line per redefinition, which is `dbp` and therefore silent at `DebugLevel == 0`.

So of my five review points, the ones that still stand are the stale-branch revert (1), the four interface versions (2), and the duplicated `WoWPro_Leveling/Forever/` guides (3). Points 4 and 5 I got wrong or overstated, and the later comment covers 5.

rmburgess-dev:
I've opened #3468, which is this PR rebased onto current `main` with fixes on top — your commits are all there, unmodified, and credited.

Two things I found while rebasing that affect this branch as pushed:

**It is five commits behind `main` and reverts #3452.** The merged result deletes `showItemButton`, `showTargetButton`, `showEAButton`, `showJumpButton` and `showLootsButtons` from the defaults, along with the guards in `WoWPro_Broker.lua` and the "Button Display Options" group in `WoWPro_Config.lua`. A rebase resolves it cleanly.

**Four of the seven Camelot TOCs declare the wrong interface** — `120007,120100` on Achievements, Leveling and Recorder, `11509` on Dailies. Forever is `16001`, and a `_Camelot.toc` that does not declare it is one Forever treats as out of date.

Also worth knowing: `WoWPro_Camelot.toc` loads `WoWPro_ZoneData_Forever.lua` but not the `WoWPro_ZoneData.lua` it overrides, and `WoWPro_WorldEvents_Camelot.toc` loads the Retail guides while the `Vanilla` tree sits unused. Details and the rest are in #3468.

No action needed from you if #3468 is the one that goes in — just did not want to land on top of your work without saying so.

PR #3458, opened 2026-09-19
Fix Recorder not saving first step on newly created guides

Spoony36:
The Recorder appeared to “not save” when users created a brand‑new guide and immediately recorded a step.

This happened because CheckpointCurrentGuide() attempted to write guide data into:

```WoWPro.Guides[currentguide]```

but for new guides, this table did not exist yet.

As a result, the first recorded step was written into nil, silently failing.

Users only noticed the issue after reloading or reopening the guide, where the first step disappeared.

This is why users had to use the workaround:

Create new guide

Hit Save manually

THEN record the first step

The manual save was the only thing that initialized the guide table.

### CODE REVIEW

File: WoWPro_Recorder/WoWPro_Recorder.lua [hunk 1]
```diff
@@ -785,7 +786,18 @@ end
 
 
 function WoWPro.Recorder:CheckpointCurrentGuide(why)
-    local GID = WoWProDB.char.currentguide
+    -- Ensure guide tables exist before saving
```
Ludovicus-Maior: Ooopsie!

Ludovicus-Maior (approved):

